PR #67, opened 2014-04-25
Add l2cap reconnection support on script part. Only Linux native part imlemented

Lahorde:
Bluetooth connection can easily be dropped. In order to support automatic reconnection :
* on linux native part : allow l2cap reconnection - default number of attempts = 5
* on script part : add 'reconnect' 'connectionDrop' events to define a reconnection policy, for an example refer to https://github.com/Lahorde/node-sensortag

TODO other platform native implementation

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,10 @@ Peripheral disconnect or cancel pending connection:
 
     peripheral.disconnect([callback(error)]);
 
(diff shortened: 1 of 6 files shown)


sandeepmistry:
Closing for now, until this can be also supported on OS X.

PR #88, opened 2014-08-26
Noble for Yosemite

DomiR:
Verified every XPC message IDs, except the one for broadcast.


OJFord:
@DomiR Seems to have changed in El Capitan, are you running the beta by any chance / do you know where to look for the changes?

PR #108, opened 2014-11-28
Added option to notify discoveries immediately

ecksun:
This option disables the buffering of broadcast data. Setting this
option will let noble emit a 'discover' event for each broadcast on
linux, instead of only emitting the event for every second broadcast
received.

The BLE sensors we use are only sending a broadcast every other minute, so we don't really want to miss every other reading.

I'm not to fond of the implementation, but this was the only way I could think of that wouldn't break backwards-compability.

I'm more than happy to refactor/rewrite it.

### DIFF
--- a/lib/linux/bindings.js
+++ b/lib/linux/bindings.js
@@ -29,10 +29,10 @@ nobleBindings.onStateChange = function(state) {
   this.emit('stateChange', state);
 };
 
-nobleBindings.startScanning = function(serviceUuids, allowDuplicates) {
+nobleBindings.startScanning = function(serviceUuids, allowDuplicates, immediateDiscovery) {
   this._scanServiceUuids = serviceUuids || [];
 
-  this._hciBle.startScanning(allowDuplicates);
+  this._hciBle.startScanning(allowDuplicates, immediateDiscovery);
 };
 
 nobleBindings.onScanStart = function() {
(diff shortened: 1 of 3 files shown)


sandeepmistry:
@ecksun checkout this commit: https://github.com/sandeepmistry/noble/commit/ee3a571aa26e39e9c39bcbebe7459710b8c11c0b

There is now an environment var. that can be set to report all advertisements.

ecksun:
Oh, sorry for the late reply.

Yeah, I guess that works.

Does this mean that you decided that the way to configure noble is through environment variables?

sandeepmistry:
@ecksun I'm hesitant to change the current API for Linux only features, so have been exposing them via environment vars for now.

It would be confusing to OS X users trying to use the ```immediateDiscovery``` parameter/API you proposed as it would not work for them, but would work on Linux.

ecksun:
Well, even NOBLE_REPORT_ALL_HCI_EVENTS still only does something on linux. As I haven't been able to test anything on OS X I don't know how OS X behaves with non-connectable advertisements. But as you say, the proper fix is to get the type of the advertisement (#145) and use that in the descision.

sandeepmistry:
Yes, agreed, there is a [separate section in the read me covering all the Linux specific](https://github.com/sandeepmistry/noble#running-on-linux) env vars.

I still need to look at #145, to see if it would help to match OS X's behaviour.

PR #130, opened 2015-02-06
fix firing callback more than once when read/writing descriptors

freehaha:
Hi,

The callback registered to listen to valueRead and valueWrite of descriptors are kept indefinitely which I believe is a typo. I've added tests and a fix. Please let me know if I'm missing something or if this behavior is intentional. Thanks.

And thanks for the brilliant module =)


sandeepmistry:
@freehaha thanks for the fix!

Agreed, looks like a typo to me.

PR #131, opened 2015-02-06
Fixed out of order packets sent on Linux

BBarash:
A pop() command was causing messages to be sent out of order. This needed to be a shift(), taking the first element of the array rather than the last.


sandeepmistry:
@BBarash thanks for this!

PR #132, opened 2015-02-14
Fix error handling and adapter binding

schnippe:
After some promising tests I suddenly had some ugly Problems with SensorTag after a reboot. It just stopped working. `node test.js` resulted in a blinking cursor on the next line. 

A fresh checkout stopped with the new message about missing privilege.  
After setting the capabilities I got the blinking cursor again. Now I ran `DEBUG=* node test.js` with the following result: 

    noble stateChange unsupported +0ms

That is strange. Before booting my BLE capable device was `hci1` and the default device was `hci0`. Now it is the other way round: default is `hci1` and `hci0` supports BLE. This was consistent across multiple boots. Probably something about Plug&Play. 

After using `NOBLE_HCI_DEVICE_ID=0  DEBUG=* node test.js` I got:

    l2cap-ble BC:xx:xx:xx:xx:xx: line = connect Transport endpoint is not connected +0ms

I traced it to `src/l2cap-ble.c`. The error handling there ignored some intermediate errors so I corrected it. The root error was 

    l2cap-ble BC:xx:xx:xx:xx:xx: line = connect (hci0): Invalid request code

After browsing kernel sources in the `net/bluetooth` tree I decided to take a look at the debug output there. I enabled dynamic kernel debug via dmesg for ble:

    echo "file net/bluetooth/hci_conn.c =flmtp" > /sys/kernel/debug/dynamic_debug/control
   and repeat with `hci_core.c`, `lcap_sock.c`, `lcap.core.c`

there it hit me: `l2cap_chan_connect` was using `hci1`.
 
    ...
    bluetooth:hci_get_route:482: 00:00:00:00:00:00 -> bc:xx:xx:xx:xx:xx
    bluetooth:hci_connect:711: hci1 dst bc:xx:xx:xx:xx:xx type 0x80
    ...

It chooses which source adapter to use by calling `hci_get_route` which is using the default adapter if no source BT-address is set which is what l2cap-ble was doing. 

Please pull my change sets.

### DIFF
--- a/lib/linux/hci-ble.js
+++ b/lib/linux/hci-ble.js
@@ -40,7 +40,6 @@ HciBle.prototype.onStdoutData = function(data) {
 
     this._buffer = this._buffer.substring(newLineIndex + 1);
 
-
     debug('line = ' + line);
 
     if ((found = line.match(/^adapterState (.*)$/))) {
@@ -60,7 +59,7 @@ HciBle.prototype.onStdoutData = function(data) {
         console.log('               [sudo] NOBLE_HCI_DEVICE_ID=x node ...');
       }
 
-        this.emit('stateChange', adapterState);
+      this.emit('stateChange', adapterState);
     } else if ((found = line.match(/^event (.*)$/))) {
       var event = found[1];
       var splitEvent = event.split(',');
--- a/lib/linux/hci-ble.js
+++ b/lib/linux/hci-ble.js
@@ -40,6 +40,7 @@ HciBle.prototype.onStdoutData = function(data) {
 
     this._buffer = this._buffer.substring(newLineIndex + 1);
 
+
     debug('line = ' + line);
 
     if ((found = line.match(/^adapterState (.*)$/))) {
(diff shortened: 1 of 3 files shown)


sandeepmistry:
@schnippe good find and fix! Adding setting the address before binding makes sense.

Before merging a few changes have to be made, options are:

1) Update regex handling in [l2cap-ble.js](https://github.com/sandeepmistry/noble/blob/master/lib/linux/l2cap-ble.js), specifically [here](https://github.com/sandeepmistry/noble/blob/master/lib/linux/l2cap-ble.js#L96).

2) Revert added/modified ```printf```'s in l2cap-ble.c

Let me know which option you prefer, I'm leaning towards option 2 now, also let me know if you need any help.

schnippe:
@sandeepmistry thx. I already made the printfs in such a way that the regex from `lcap2-ble.js`capures and puts the message into status and after that  into `new Error(status)`. 

I works for me with noble-sensortag/test.js on linux. Did I miss something else?

sandeepmistry:
@schnippe I haven't tried out you changes, but won't they cause multiple ```connect``` events to fire?

The version in master expects:
 * ```connect success``` - emits connect with no error
 * ```connect <some error message>``` - emits connect with error

Only one connect event should be fired.

schnippe:
@sandeepmistry I have a `goto done;` in every error printf in `lcap2-ble.c`. `done` does a `printf("disconnect\n");` and closes open handles. So there will be only one `printf ("connect ...` in either case. Only one connect event will fire .

sandeepmistry:
@schnippe thanks for the explanation, I missed the ```goto```'s when reviewing earlier.

schnippe:
@sandeepmistry Thank you!

PR #145, opened 2015-03-07
Detect type of event of the advertising info

doganyazar:
This addresses adding the capability to detect the type of the advertising info discussed in the issue: "Added option to notify discoveries immediately #108"


sandeepmistry:
Noble now uses [bluetooth-hci-socket](https://github.com/sandeepmistry/node-bluetooth-hci-socket) instead of spawned (```hci-ble``` and ```l2cap-ble```) processes.

A similar change has been incorporated.

Thanks for the pull request though!

Issue #150, opened 2015-03-20
`Uncaught TypeError: Cannot read property 'uuid' of undefined` error


taktran:
I randomly get this error on OSX 10.8.5, noble v0.3.10:

```
Uncaught TypeError: Cannot read property 'uuid' of undefined
    at eval (/Users/.../node_modules/noble/lib/mac/legacy.js:392:59)
    at EventEmitter.emit (events.js:95:17)
    at eval (/Users/.../node_modules/noble/lib/mac/legacy.js:44:8)
    at EventEmitter.emit (events.js:95:17)
    at eval (/Users/.../node_modules/noble/lib/mac/legacy.js:20:10)
    at XpcConnection.EventEmitter.emit (events.js:95:17)
```

Looking at the code where the error is generated:

```
nobleBindings.on(isLessThan10_8_5 ? 'kCBMsgId35' : 'kCBMsgId56', function(args) {
  var peripheralHandle = args.kCBMsgArgPeripheralHandle;
  var peripheralUuid = this._peripherals[peripheralHandle].uuid;
  var characteristicHandle = args.kCBMsgArgCharacteristicHandle;
  var isNotification = args.kCBMsgArgIsNotification ? true : false;
  var data = args.kCBMsgArgData;

  for(var i in this._peripherals[peripheralHandle].services) {
    if (this._peripherals[peripheralHandle].services[i].characteristics &&
        this._peripherals[peripheralHandle].services[i].characteristics[characteristicHandle]) {

      this.emit('read', peripheralUuid, this._peripherals[peripheralHandle].services[i].uuid,
        this._peripherals[peripheralHandle].services[i].characteristics[characteristicHandle].uuid, data, isNotification);
      break;
    }
  }
});
```

https://github.com/sandeepmistry/noble/blob/v0.3.10/lib%2Fmac%2Flegacy.js#L392

It fails on `this._peripherals[peripheralHandle]` being undefined.

Here are some values I got when I was console logging:

```
args = {"kCBMsgArgData":{"0":86,"length":1,"parent":{"0":86,"length":1},"offset":0},"kCBMsgArgIsNotification":1,"kCBMsgArgCharacteristicHandle":25,"kCBMsgArgPeripheralHandle":13}
peripheralHandle = 13
this._peripherals[peripheralHandle] = {}", 
```

```
[...]

PR #159, opened 2015-04-05
Bug of undefined peripheral on legacy mac

taktran:
Fixes https://github.com/sandeepmistry/noble/issues/150, where peripheral is undefined for legacy macs.


sandeepmistry:
@taktran thanks! I'll port to Mavericks and Yosemite files as well...

PR #170, opened 2015-04-17
Update uuid-to-address.js

petergaultney:
this minor change seems to have fixed a case on my mac where CoreBluetoothCache was not a defined member of obj[0].

### DIFF
--- a/lib/mac/uuid-to-address.js
+++ b/lib/mac/uuid-to-address.js
@@ -16,7 +16,7 @@ module.exports = function(uuid, callback) {
                         uuid.substring(16, 20) + '-' +
                         uuid.substring(20);
 
-    var coreBluetoothCacheEntry = obj[0].CoreBluetoothCache[formattedUuid];
+    var coreBluetoothCacheEntry = obj[0].CoreBluetoothCache ? obj[0].CoreBluetoothCache[formattedUuid] : undefined;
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@petergaultney thanks!

I though I had fixed the issue with: https://github.com/sandeepmistry/noble/commit/b6ff96b44f7627bee4b055cee77dbbf01c201b07 - is that not the case?

petergaultney:
I did see that recent commit, but it doesn't seem to have fixed it, no. Basically, CoreBluetoothCache itself wasn't defined, so it never got to the next line that had your ternary fix on it.

petergaultney:
Sandeep, my apologies - I had just installed from npm, and so I didn't even notice that the version in npm didn't have that fix, while your repo does. 

I haven't verified it, because I'm totally new to Node.js and npm and not totally sure how to go about pulling in the most recent changes, but it looks to me like your solution should work, assuming it lets the connection proceed despite creating the Error object. Mine, in any case, actually successfully connected to the BLE slave device when I simply returned "callback(null, null);"

In any case, I really appreciate your library. It's made my evening a whole lot more successful.

sandeepmistry:
@petergaultney no worries, it is confusing that master and npm weren't in sync.

I published v0.3.14 yesterday which includes the fix from https://github.com/sandeepmistry/noble/commit/b6ff96b44f7627bee4b055cee77dbbf01c201b07

Issue #85, opened 2014-07-31
TypeError: Cannot read property 'services' of undefined


bmiller59:
Hello. I am running in Mavericks trying to connect my sensortag as a node red module. I have enabled DEBUG=mavericks-bindings and I can see that it discovers and connect correctly and that there is a variety of message activity and then the following. The last debug message I added in the msg70 handler. It appears that the _peripherals object is empty for some reason, leading to the services property to be undefined. Do you know why? Thank you.

 "kCBMsgId": 70
} +1ms
  mavericks-bindings xpcEvent: {
  "kCBMsgArgs": {
    "kCBMsgArgData": [
      216,
      111,
      62,
      129
    ],
    "kCBMsgArgIsNotification": 1,
    "kCBMsgArgCharacteristicHandle": 55,
    "kCBMsgArgDeviceUUID": [
      240,
      167,
      253,
      23,
      136,
      137,
      77,
      233,
      143,
      163,
      91,
      130,
      89,
      49,
      218,
      182
    ]
  },
  "kCBMsgId": 70
} +36s
  mavericks-bindings Msg70 +1ms f0a7fd1788894de98fa35b825931dab6 {} undefined

30 Jul 21:34:32 - [red] Uncaught Exception:
30 Jul 21:34:32 - TypeError: Cannot read property 'services' of undefined
    at nobleBindings.write.sendCBMsg.kCBMsgArgDeviceUUID (/Users/brendan_miller/projects/sensortag/node-red/node_modules/noble/lib/mac/mavericks.js:375:45)
    at emit (events.js:95:17)
    at nobleBindings.sendCBMsg.sendXpcMessage.kCBMsgId (/Users/brendan_miller/projects/sensortag/node-red/node_modules/noble/lib/mac/mavericks.js:41:8)
    at emit (events.js:95:17)
    at null.<anonymous> (/Users/brendan_miller/projects/sensortag/node-red/node_modules/noble/lib/mac/mavericks.js:17:10)
    at XpcConnection.emit (events.js:95:17)

vielmetti:
I'm seeing what I think is this very same error (OS X Mavericks), investigating.

FluffyJack:
Seeing this error on yosemite. Seems like it's loosing the data some time during the connection. It happens only every few messages.

```
/Users/fluffyjack/Developer/mini_drone/node_modules/noble/lib/mac/yosemite.js:525
[...]

PR #171, opened 2015-04-20
persistent peripherals between calls to .startScanning() on mavericks

andySigler:
periphal objects stored in `_peripherals` are no longer overwritten when repeated picked up while scanning


sandeepmistry:
@andySigler thanks! This address #85 on OS X 10.9

Issue #176, opened 2015-04-30
noble warns BLE state is unsupported state sporadically


LookingSharp:
When I register for state change I occasionally get `unsupported` for the state.  I'm not sure why this happens, but to get the expected `poweredOn` state I need to kill and restart the app.  Is there a way to recover from this from within the program?  A way to tell noble to reset?

```
noble warning: adapter does not support Bluetooth Low Energy (BLE, Bluetooth Smart).
               Try to run with environment variable:
               [sudo] NOBLE_HCI_DEVICE_ID=x node ...
ble: stateChange: state = unsupported
```

I'm running on a Raspberry Pi.

Thanks.

taktran:
I'm also getting this. I was thinking of running the app on a background thread and restarting it for now. But a restart method of some sort would be amazing!

haughank:
I have seen the same error on a Fujitsu Lifebook A Series running ubuntu 14.04 LTS. Doesn't seem to be a Raspberry Pi only issue. Again, restarting the application solves the problem

haughank:
Additional note: the frequency of this occurring is about 20% of the time that the app is launched

sandeepmistry:
@LookingSharp @taktran @haughank if you could provide some steps to reproduce that would be great. As well as the output of ``hcidump``.

Which BLE USB adapters are you using?

LookingSharp:
It is intermittent, so there aren't good repro steps.  Just getting the state change, which is the first thing the program does, causes the problem.

I'm trying to capture it in hcidump, but thus far I haven't had luck.  It isn't reproing for me at the moment.

I'm using the IOGear adapter
http://www.amazon.com/IOGEAR-Bluetooth-Micro-Adapter-GBU521/dp/B007GFX0PY/

LookingSharp:
Is there any way to work around this node?  Is there a way to shut down nobel and restart it from within Node?

LookingSharp:
@sandeepmistry This is the output for hcidump -

```
pi@raspberrypi ~/node $ hcidump
HCI sniffer - Bluetooth packet analyzer ver 5.30
device: hci0 snap_len: 1500 filter: 0xffffffff
> HCI Event: Command Complete (0x0e) plen 4
    LE Set Scan Parameters (0x08|0x000b) ncmd 1
    status 0x0c
    Error: Command Disallowed
```

[...]

PR #182, opened 2015-05-12
scanStart does not emit an error parameter

LookingSharp:


sandeepmistry:
@LookingSharp I've made that change recently, it hasn't been published to npm yet.

See: https://github.com/sandeepmistry/noble/blob/master/lib/noble.js#L69

Let me know if I'm missing something else though ...

LookingSharp:
@sandeepmistry that makes sense, but this fix is actually going to make things harder for me.

[Issue 176](https://github.com/sandeepmistry/noble/issues/176) tracks the case where sometimes the adapter comes back unsupported, even though it may have worked on the run before and will work on the next run.  Recent investigation revealed that if I ignore that and start scanning anyway, things still work (though noble.state never recovers from unsupported).  With this fix I won't be able to scan in this case.  Let me know if you get a chance to look at the new info in issue 176.

sandeepmistry:
@LookingSharp I'm confident #176 will be solved before publishing a new version to npm.

LookingSharp:
Awesome, thanks @sandeepmistry

PR #189, opened 2015-06-01
Add packages needed for rpm-based systems

ppannuto:
Tested only on my local fedora 21, but probably general.


sandeepmistry:
@ppannuto thanks!

PR #240, opened 2015-09-21
fix this._discoveredPeripheralUUids = []; variable not initalized in constructor

jacobrosenthal:
Clearly not giving you trouble currently, but when I copied this code and ran it in react native it came up, so why not fix it.

### DIFF
--- a/lib/noble.js
+++ b/lib/noble.js
@@ -33,6 +33,7 @@ function Noble() {
   this._services = {};
   this._characteristics = {};
   this._descriptors = {};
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@jacobrosenthal sure, why not.

Curious to know what issues you were seeing with react native, without it ...

jacobrosenthal:
If you're really curious I ll break it again, but it would be set by
ondiscover but was undefined when whatever other function went to use it.

PR #251, opened 2015-10-15
Fix a potential exception when accessing a non existent element

Loghorn:
It could happen (and it sometime does on my machine) that an the
_discoveries map does not contain the address when the type is 0x04


sandeepmistry:
@Loghorn thanks!

PR #278, opened 2015-12-02
Pass notification state to characteristic notify callback

rclai:
The notification state is being emitted by the `Noble` object [here](https://github.com/sandeepmistry/noble/blob/a7837885e520578a64294e249e44a96f891624f1/lib/noble.js#L326), but it is not sent through to the `notify` callback.

### DIFF
--- a/lib/characteristic.js
+++ b/lib/characteristic.js
@@ -87,8 +87,8 @@ Characteristic.prototype.broadcast = function(broadcast, callback) {
 
 Characteristic.prototype.notify = function(notify, callback) {
   if (callback) {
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@rclai what's the use case for this? Typically set operations don't return/callback with the value passed in.

rclai:
The use case is to make sure that the notify did actually get set. In iOS, there is a delegate that gets called when set a notify and you're able to see in there the actual notify state of the characteristic.

jacobrosenthal:
This is deprecated by the .subscribe and .unsubscribe api

Issue #175, opened 2015-04-29
peripherals with duplicate services or characteristics are not handled properly


corecode:
All services and characteristics are accessed via hashes.  If a peripheral contains multiple services/characteristics of the same uuid, only the last discovered service/characteristic is accessible.

A possible solution would be to carry the handle to the objects, and perform accesses only using handles.

OJFord:
Same UUID characteristics in different services work fine for me. (Maybe this has been fixed but not closed?)

Are same-UUID services, or same-UUID characteristics within a service even legal?

sandeepmistry:
@OJFord it's still an issue.

Yes, a peripheral can have one or many services with the same UUID, and one or many characteristics with the same UUID in a sever.

VinceGa:
Yes, this is a very legit request. Currently we could read/write attributes via handles. But there's no way(not AFAIK) to get the handle of a specific attribute. We should add this property in the discover/read/write APIs.

sandeepmistry:
@VinceGa yes, it's on my TODO list, plan is to add an ```id``` field to all attribute objects (services, characteristics, descriptors). For most platforms noble supports it will be mapped to the attribute handle, however if that info is not available another unique number could be used.

P.S. reading attributes by handle works well on Linux, however on OS X CoreBluetooth/blued blocks this operation sometimes.

aiham:
I've run into this problem too. I cannot listen to notify events by a characteristic with the same UUID as another characteristic.

LarsBuur:
@sandeepmistry you wrote above: "P.S. reading attributes by handle works well on Linux, however on OS X CoreBluetooth/blued blocks this operation sometimes." 

How can I detect that the operation is blocked?

rzr:
Please can you update status of this issue on this latest release: 

https://www.npmjs.com/package/@abandonware/noble/v/1.9.2-12

OJFord:
@rzr I have no idea because I haven't used this project in years, but it looks like that package is being maintained (and is actually being maintained, not just hosted) at abandonware/noble, so you'll probably have better luck checking for a similar (perhaps closed) issue there, or asking via Discussions or something.

PR #290, opened 2015-12-19
Read handles from characteristics and services for linux platform

janoelker:
As the topic sais, I changed the linux part of the system, to make it able to read the handle for a characteristic or service. These values are needed in future use for me, so I would like to hear your opinion on this commit. 

Also is it correct to return the startHandle or is a combination of the startHandle and endHandle needed?

### DIFF
--- a/lib/characteristic.js
+++ b/lib/characteristic.js
@@ -5,7 +5,7 @@ var util = require('util');
 
 var characteristics = require('./characteristics.json');
 
-function Characteristic(noble, peripheralId, serviceUuid, uuid, properties) {
+function Characteristic(noble, peripheralId, serviceUuid, uuid, properties, handle) {
   this._noble = noble;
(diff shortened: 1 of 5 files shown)


sandeepmistry:
@Steckdoserich thanks for submitting this PR!

I'm not too keen on promoting the use of handles. However, I am planning on introducing an ```id``` field for services, characteristics, and descriptors to resolve #175. The id will most likely be mapped to the service start handle, and characteristic/descriptor handle for both OS X and Linux.

tal:
@sandeepmistry I'm actually hoping to have handle available for debugging purposes. I know the handle based on sniffing traffic on a device and I'm trying to be able to explore the interface based on that. So having it accessible would be very useful.

rzr:
#923

rzr:

PR #291, opened 2015-12-21
Added logic to reject rather than ignore unknown requests/commands.

george-hawkins:
This is a simplified version of pull request [`#277`](https://github.com/sandeepmistry/noble/pull/277) as per Sandeep's [comment of Dec 14th](https://github.com/sandeepmistry/noble/pull/277#issuecomment-164313393).


sandeepmistry:
@george-hawkins thanks!

PR #297, opened 2016-01-05
fix serviceUuids var check in peripheral-explorer.js

jrobeson:
it was previously using localName to check.


sandeepmistry:
@jrobeson thanks!

Issue #300, opened 2016-01-11
Way to get callback if some other process changes scan parameters?


bradjc:
I'm running into an issue where I'm trying to have a long lived noble script scan continuously for BLE advertisements, but if anything else on the system uses the BLE/HCI subsystem it can change the scanning parameters unbeknownst to the noble script. For example, running

    sudo hcitool lescan

while the noble script is running causes the noble script to only receive unique advertisements. Ending `lescan` stops all advertisements.

I think the best way to address this is to augment `Gap.prototype.onHciLeScanEnableSet` ([here](https://github.com/sandeepmistry/noble/blob/master/lib/hci-socket/gap.js#L40)) with an `else` statment that emits a callback if the scan parameters changed outside of the script. Then the script can change them back.

Ideally this wouldn't be an issue, but it's hard to make a robust application if there is no way to detect conflicts like these.

sandeepmistry:
noble might be able to listen for HCI commands (filter needs to be changed), and react based on the status.

* Here is where the filter is currently set: https://github.com/sandeepmistry/noble/blob/master/lib/hci-socket/hci.js#L117
* Here is where socket data is processed: https://github.com/sandeepmistry/noble/blob/master/lib/hci-socket/hci.js#L371

> I think the best way to address this is to augment Gap.prototype.onHciLeScanEnableSet (here) with an else statment that emits a callback if the scan parameters changed outside of the script. Then the script can change them back.

That works for me.

sandeepmistry:
Closed via #301.

PR #301, opened 2016-01-19
Check for changes to scanning

bradjc:
This adds receiving HCI CMD packets and checking to see if they are scan enable commands. If they are, it checks to see if scan enable/disable and duplicate filtering settings are different then what the current script expects. If they are, a callback is issued so that the main script can take any necessary action.

Pull request for issue #300.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,11 +232,17 @@ noble.on('stateChange', callback(state));
 
 #### Scan started:
 
+This callback is called once scanning has started. It can also be called
+if some other application enables BLE scanning.
+
 ```javascript
 noble.on('scanStart', callback);
 ```
 
 #### Scan stopped
+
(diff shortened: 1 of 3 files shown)


sandeepmistry:
@bradjc thanks for submitting!

I'd prefer not to add a new Linux specific event to noble. What do you think about re-using the existing ```scanStart``` and ```scanStop``` events?

bradjc:
I updated the PR to use the existing callbacks.

sandeepmistry:
@bradjc thanks!

I've merged and made a few changes (https://github.com/sandeepmistry/noble/commit/c81097be7fdaf532cc4047e76bcc97823a0c4c7a). The filter duplicates values is now trickled up to the ```scanStart``` event.

PR #312, opened 2016-01-26
Add parameters to scan and connect

Loghorn:
It could be useful to allow the user to change scan and connect parameters

### DIFF
--- a/lib/hci-socket/bindings.js
+++ b/lib/hci-socket/bindings.js
@@ -30,6 +30,10 @@ var NobleBindings = function() {
 util.inherits(NobleBindings, events.EventEmitter);
 
 
+NobleBindings.prototype.setScanParameters = function(interval, window) {
+  this._gap.setScanParameters(interval, window);
+}
+
 NobleBindings.prototype.startScanning = function(serviceUuids, allowDuplicates) {
   this._scanServiceUuids = serviceUuids || [];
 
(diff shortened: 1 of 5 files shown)


sandeepmistry:
@Loghorn thanks for submitting! Changes look good.

However, I am concerned about the new functionality not being available on OS X.

Loghorn:
Unfortunately I don't have a Mac to test the changes, but why do you have concerns? With the changes I made you can only pass different parameters from the ones you choose by default, but the sequence of calls is the same. Maybe some value combinations could not work on OS X, but this is also true for Linux or Win, so the user must check if they are valid before making the calls; or she can choose to use your defaults and don't change the parameters and noble will work as before.

sandeepmistry:
@Loghorn OS X does not use the HCI socket based code, it interacts with blued/CoreBluetooth.

So only the functionality provided by the CoreBluetooth API is available: https://developer.apple.com/library/mac/documentation/CoreBluetooth/Reference/CoreBluetooth_Framework/

My concern with the new API's are an app using the new features would behave differently on OS X than Linux or Windows.

Loghorn:
@sandeepmistry  I see. My experience with OS X is very limited so I cannot make any changes to how noble works on it confidently enough.
However, my PR consist of two parts:
   1. discovery parameters
   2. `connect` parameters

After looking at the documentation you linked, I found no way to change discovery parameters on OS X, so the `setScanParameters` function could simply `emit('scanParametersSet')` (maybe with a `false` argument to propagate or something?)
For the connection parameters, on OS X you can only set connection latency with Peripheral Manager's `setDesiredConnectionLatency`. It's not clear to me if it applies the latency to all subsequent connections or also to existing connections: if the latter is true there is nothing to do, but if the former is true we can change the `connect` function to accept the latency setting. Or the `connect` parameters can simply be ignored by OS X: a node application using noble will still be able to connect to a Bluetooth device when running on OS X even if the connection parameters could not be customized. It's not much different from how it works now because you choose arbitrary parameters for `connect` on non-OS X system and this could behave differently if they are not the same as OS X ones (and even if you set them identical to OS X, no one can prevent Apple from changing them since they are not exposed in the API)

sandeepmistry:
> For the connection parameters, on OS X you can only set connection latency with Peripheral Manager's setDesiredConnectionLatency.

Yes, unfortunately the Peripheral Manager API cannot be used, it is for peripheral mode and not central mode (see Central Manager).

rzr:
#923

rzr:
Please also check with:

Issue #313, opened 2016-01-26
Double ServiceUUID bug


paulgavrikov:
Hi Sandeep,

I have been using you lib for 2 months now and added bunch of functions (hci / linux only). I also found a bug: When using OS X devices, some apps push services like '180a' which are already provided by OS X or iOS Gatt. Which means that if one discovers services he will get that service twice. Note that those services might have totally different characteristics. Since noble is not using handles, discovering Characteristics of a Services that is non-unique will only discover characteristics of one of those non-unique services.

I totally agree that this is against the BLE Spec, but it happens occasionally. E.g. when launching virtual Peripherals from the LightBlue iOS app.

This might be a minor problem but also causes a bug where ```discoverAllServicesAndCharacteristics ``` or ``` disocverSomeServicesAndCharacteristics``` will get stuck. 

I patched this in ```noble/lib/hci/peripheral.js``` but I don't know how to create a pull request so I post it here. Note: this only helps the function to terminate, so it's a workaround for off-spec BLE Devices.

```
Peripheral.prototype.discoverSomeServicesAndCharacteristics = function(serviceUuids, characteristicsUuids, callback) {
  this.discoverServices(serviceUuids, function(err, services) {


    var servicesLength = services.length;
    var knownUUIDs = [];

    for (var s in services) { 
      var service = services[s];
      if(knownUUIDs.indexOf(service.uuid) != -1)
        servicesLength--;
      knownUUIDs = knownUUIDs.concat(service.uuid);
    }

    var numDiscovered = 0;
    var allCharacteristics = [];


    for (var i in services) {
      var service = services[i];

      service.discoverCharacteristics(characteristicsUuids, function(error, characteristics) {
        

        numDiscovered++;

        if (error === null) {
          for (var j in characteristics) {
            var characteristic = characteristics[j];

            allCharacteristics.push(characteristic);
          }
        } 

        if (numDiscovered === servicesLength) {
          if (callback) {
            callback(null, services, allCharacteristics);
          }
        }
      }.bind(this));
    }
  }.bind(this));
};
```

jrobeson:
creating a pull request is easy in the web editor. Just navigate to the file and click the edit button.

sandeepmistry:
@paulgavrikov thanks for reporting.

> I totally agree that this is against the BLE Spec, but it happens occasionally. E.g. when launching virtual Peripherals from the LightBlue iOS app.

Actually it's allowed, I think it would be better to fix #175 than working around it.

[...]

PR #316, opened 2016-01-27
(mostly) replace peripheralUuid with peripheralId

jrobeson:
This doesn't include changes to websocket related calls or events since that
would seem to break the public API

There are some internal peripheral.uuid references that should be looked at though.

### DIFF
--- a/lib/distributed/bindings.js
+++ b/lib/distributed/bindings.js
@@ -157,8 +157,8 @@ NobleBindings.prototype.stopScanning = function() {
   this.emit('scanStop');
 };
 
-NobleBindings.prototype.connect = function(deviceUuid) {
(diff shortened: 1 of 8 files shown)


jrobeson:
If you were to remove distributed and websocket bindings as mentioned in https://github.com/sandeepmistry/noble/issues/313#issuecomment-177288884, then this would be all done except in hci-socket I think (unless you disagree with some of the changes internal to the mac bindings)

jrobeson:
ping?

jrobeson:
how time flies.

jacobrosenthal:
Related to #597, I think were going to look at refactoring ids in 2.0 instead of rewriting 1.0

PR #318, opened 2016-01-27
Update peripheral.js

paulgavrikov:
Fixes a bug, where discovery would never exit if a service UUID is non unique.


jrobeson:
there are some unnecessary whitespace changes in there. Those are gonna have to be removed

jacobrosenthal:
Well get to this in a 2.0 with unique ids, which should support duplicate uuids

PR #319, opened 2016-01-27
Reopen the socket when in bad state

Loghorn:
On my machine, when the bluetooth driver is restarted, noble does not handle it because the underlying socket is in a bad state. This fix will try to reopen the socket under those circumstances.

### DIFF
--- a/lib/hci-socket/hci.js
+++ b/lib/hci-socket/hci.js
@@ -67,9 +67,10 @@ var HCI_OE_USER_ENDED_CONNECTION = 0x13;
 var STATUS_MAPPER = require('./hci-status');
 
 var Hci = function() {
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@Loghorn 

> when the bluetooth driver is restarted

Can you please provide more detail on this? (which OS and adapter, how/why is it being restarted).

A state change of "resetting" should probably be triggered, when the "File descriptor in bad state" is detected.

Loghorn:
I tested it on Linux, with a AP6335 chip. The driver is restarted by manually reloading the firmware on the chip (I've to do that because sometimes it hangs :disappointed: )
However, after further testing it, the fix is not working as intended and some cases are not covered, so I'm going to close the PR and will reopen it if I'll implement a proper fix.

PR #320, opened 2016-01-29
Accept addresses in peripheral-explorer.js

jrobeson:


sandeepmistry:
@jrobeson good idea, however I think it would be clearer if the .address property is checked.

```javascript
if (peripheral.id === peripheralId || peripheral.address === peripheralId) {
  // ...
}
```

I'm open to suggestions on a better name for the ```peripheralId``` var ...

jrobeson:
I thought you wanted it to accept peripheralId. If you really want it to accept addresses only, then I will certainly check against address only.

sandeepmistry:
@jrobeson I was thinking we could support passing in a peripheral id or BT address as a CLI arg.

Then check both ```peripheral.id``` and ```peripheral.address``` for matches. Instead of removing semicolons from the CLI arg.

jrobeson:
updated and rebased

sandeepmistry:
@jrobeson thanks!

I've pushed an additional change to rename ```peripheralId``` to ```peripheralIdOrAddress``` (https://github.com/sandeepmistry/noble/commit/05aa8934736e554ee553709e95aadf2d2728d72b).

PR #342, opened 2016-02-28
Add a .travis.yml

jrobeson:


jrobeson:
once nvm is natively supported in osx, we can go back to the much cleaner implementation

sandeepmistry:
First travis build has passed thanks! I'll add a badge to the read me.

PR #238, opened 2015-09-15
Web Bluetooth API shim for noble.

monteslu:
## WebBluetooth API is now ready enough to use for several types of BLE apps.

The intent of this is to be able to webpack/browserify noble without the need for a websocket server or other plugins.  The bindings should target this spec:  https://webbluetoothcg.github.io/web-bluetooth/

A few notes:
- I've only gotten web bluetooth to connect on ChromeOS behind a flag:
  chrome-flags
- This code first tries for web bluetooth api, then falls back to the websocket bindings.

cc:// @sandeepmistry @jacobrosenthal @beaufortfrancois

PR #366, opened 2016-03-28
Support for pluggable bindings

hgwood:
Some weeks ago I wrote basic Universal Windows Platform bindings for Noble. These bindings are available at https://github.com/hgwood/winble. They are still pretty raw.

While developing those bindings I found out there is no easy way to make Noble use bindings that it is not aware of. Basically you have to edit `noble.js`. I think it would be useful if Noble users (especially bindings developers) could optionally load any bindings they want and make Noble use them without editing its source.

This PR is an attempt to bring this feature to Noble without breaking the API. As of yet it is just a quick sketch of what I'm thinking of and I haven't tested it. Hopefully, from there, we can start a discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,12 @@ npm install noble
 var noble = require('noble');
 ```
 
+By default, Noble will select bindings to communicate with Bluetooth devices depending on your platform. If you prefer to specify what bindings Noble should use
+
+```javascript
+var noble = require('noble/with-bindings')(require('./my-custom-bindings'));
+```
+
 ### Actions
 
 #### Start scanning
--- a/lib/noble.js
+++ b/lib/noble.js
@@ -9,27 +9,26 @@ var Service = require('./service');
 var Characteristic = require('./characteristic');
 var Descriptor = require('./descriptor');
 
-var bindings = null;
-
-var platform = os.platform();
-
-if (process.env.NOBLE_WEBSOCKET || process.title === 'browser') {
-  bindings = require('./websocket/bindings');
-} else if (process.env.NOBLE_DISTRIBUTED) {
-  bindings = require('./distributed/bindings');
-} else if (platform === 'darwin') {
-  bindings = require('./mac/bindings');
-} else if (platform === 'linux' || platform === 'win32') {
-  bindings = require('./hci-socket/bindings');
-} else {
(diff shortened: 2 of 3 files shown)


sandeepmistry:
@hgwood cool!

@monteslu has done something similar in PR #238 to split out the binding resolver: https://github.com/sandeepmistry/noble/pull/238/files#diff-a3c59ae9a2f115e917fefe6609846672R1

noble has been using environment vars to control non-API flags, so another option would be to specific the binding via env var, so something like this:

```js
process.env.NOBLE_BINDING = 'blah';
var noble = require('noble');
```

or command line
```
NOBLE_BINDING=blah node <file>.js
```

Thoughts? I'm ok when the ```require('noble/with-bindings')``` you suggested.

hgwood:
The `with-binding` solution is more flexible because it allows to pass an arbitrary object as bindings and that is useful when developing those, whereas with the environnement variable the bindings have to be in an installed module. The environment variable is still a useful possibility though, and I can add support for it.

hgwood:
Oh, wait. I guess `NOBLE_BINDING` could be set to something like `./my-bindings` and that would work also when developing bindings. If you find this technique more in line with how noble works I'm OK with using it. It has the advantage that the code change required is a lot less significant. I'll update the PR later.

sandeepmistry:
@hgwood your feedback from https://github.com/sandeepmistry/noble/pull/366#issuecomment-206202139 has convinced me to go with your original patch :) even though it is a slightly bigger code change. I'll merge later this week, I also want to split the "guessBindings" part out. As well, as start an advance usage section in the read or separate markdown file.

hgwood:
OK then. Ask me if you need any help.

sandeepmistry:
@hgwood thanks!

Done, I've merged with the changes I mentioned in https://github.com/sandeepmistry/noble/pull/366#issuecomment-206634022, see https://github.com/sandeepmistry/noble/commit/e0b48b58f875979d536629063df29bba5544ed11 and https://github.com/sandeepmistry/noble/commit/c1e7a065f24fa0530bde41a1efab8b4917eb2965.

You can try out via ```npm install sandeepmistry/noble``` prior to the next publish to npm.

hgwood:
Great!

PR #367, opened 2016-03-29
Don't kill all descriptors when looking for new Characteristics

Neutrosider:
Previously when looking for characteristics, all descriptors of the
related service were deleted. This caused a bug when messages on one of
the descriptors arrived before the descriptors were rediscovered. This
should fix it.

Please consider what consequences this change might have, as i don't have in-depth knowledge about noble, **and don't know if this might cause any new bugs**.

Fixed the following Bug for me:

`/root/node/node_modules/noble/lib/hci-socket/gatt.js:608`
`        this._descriptors[serviceUuid][characteristicUuid][descriptors[i].uuid] = descriptors[i];`
`                                                                                ^`

`TypeError: Cannot set property '2902' of undefined`
`    at null.<anonymous> (/root/node/node_modules/noble/lib/hci-socket/gatt.js:608:81)`
`    at Gatt.onAclStreamData (/root/node/node_modules/noble/lib/hci-socket/gatt.js:122:26)`
`    at emitTwo (events.js:92:20)`
`    at emit (events.js:172:7)`
`    at AclStream.push (/root/node/node_modules/noble/lib/hci-socket/acl-stream.js:35:10)`
`    at NobleBindings.onAclDataPkt (/root/node/node_modules/noble/lib/hci-socket/bindings.js:263:15)`
`    at emitThree (events.js:97:13)`
`    at emit (events.js:175:7)`
`    at Hci.onSocketData (/root/node/node_modules/noble/lib/hci-socket/hci.js:447:14)`
 `   at emitOne (events.js:77:13)`

### DIFF
--- a/lib/hci-socket/gatt.js
+++ b/lib/hci-socket/gatt.js
@@ -380,7 +380,9 @@ Gatt.prototype.discoverCharacteristics = function(serviceUuid, characteristicUui
   var characteristics = [];
 
   this._characteristics[serviceUuid] = {};
-  this._descriptors[serviceUuid] = {};
+  
+  if (this._descriptors[serviceUuid] == null)
+	this._descriptors[serviceUuid] = {};
 
   var callback = function(data) {
     var opcode = data[0];


sandeepmistry:
@Neutrosider thank you for submitting!

Your patch is causing the Travis CI build to fail JSHint linting.

Could you please share a code snippet on what sequence this patch solves?

Neutrosider:
Sure, but can it not be seen in the "files Changed"-tab? if not, i only added a single if-statement. in lib/hci-socket/gatt.js, line 383, i replaced this line:
```Javascript
this._descriptors[serviceUuid] = {};
```
with these two:
```Javascript
if (this._descriptors[serviceUuid] == null)
    this._descriptors[serviceUuid] = {};
```

sandeepmistry:
To clarify, I'm looking for a code example of how to re-create the issue the patch resolves.

So, I can test with and without the patch.

Neutrosider:
Ah sorry, i misunderstood. This is my first pull-request ^^ It's currently 23:35 where i am, but i'll give you a test-code when i'm at work tomorrow. When i remember correctly, the easiest way to trigger this is to just constantly scan for new characters while beeing connected to a device that sends notifications. When a message arrives just after a scan started (so descriptors are cleared), but before the characteristics where rediscovered, the javascript-error happens

Neutrosider:
I made a small nodejs-script, that triggers the bug. Just place the [bug.js](https://gist.github.com/Neutrosider/02e143816b273f1c4d4d0bbcc6f27468) somwhere, where `require('noble')` wont fail, and run it via `node bug.js`
The script will start connecting to everything it can find, and enable notifications for every service that supports them. It will then continously rescan for characteristics. I have quite a few BLE devices aorund here, and without the fix, this script crashes for me within about 30 seconds (sometimes it just takes 5 seconds). With the fix, the script just keeps on running

sandeepmistry:
@Neutrosider got it, thanks for this. I've merged with a small change (https://github.com/sandeepmistry/noble/commit/59cc0cee907a894884ee166bf57dee4ee0ee8faa).

You can try it out via ```npm install sandeepmistry/noble```

Neutrosider:
Ok, i tested it with the [bug.js](https://gist.github.com/Neutrosider/02e143816b273f1c4d4d0bbcc6f27468) i provided you, and it seems to work just fine. Thanks :)

sandeepmistry:
@Neutrosider great, thanks for verifying!

PR #369, opened 2016-03-30
Add Noble constructor parameter as an option to select adapter

manuelmitasch:
Add parameter deviceId to Noble constructor as a second option to select the bluetooth adapter.
Constructor parameter has precedence over environment variable.

Implemented for hci-socket only.

### DIFF
--- a/lib/hci-socket/bindings.js
+++ b/lib/hci-socket/bindings.js
@@ -61,7 +61,7 @@ NobleBindings.prototype.updateRssi = function(peripheralUuid) {
   this._hci.readRssi(this._handles[peripheralUuid]);
 };
 
-NobleBindings.prototype.init = function() {
+NobleBindings.prototype.init = function(deviceId) {
   this.onSigIntBinded = this.onSigInt.bind(this);
 
   process.on('SIGINT', this.onSigIntBinded);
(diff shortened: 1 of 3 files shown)


sandeepmistry:
@manuelmitasch thank you for submitting.

Could you please provide a code sample of how you would use your changes?

manuelmitasch:
The end goal is to be able to use multiple bluetooth dongles. For each dongle one would create a separate Noble instance. I have identified and fixed some issues in my local branch before that will properly work. It needs some more work, but I'm very busy for the next week. 

I thought this PR is a first step in the right direction and get some feedback from you. As it does not change any "public" APIs, it is quite low risk. If somebody wants to use this feature he would need to import 'noble/lib/noble' instead of just 'noble'.

Here an example:

```
var Noble = require('noble/lib/noble');
var deviceId = 0;
var noble = new Noble(deviceId);
```

brettdaman:
Is it possible to include this in the next update, It would simplify my implementation.

sandeepmistry:
> I thought this PR is a first step in the right direction and get some feedback from you. As it does not change any "public" APIs, it is quite low risk. If somebody wants to use this feature he would need to import 'noble/lib/noble' instead of just 'noble'.

Yes, it's low risk, but does expose noble internals, which could change at anytime.

Now that #366 is merged, the change to noble.js isn't needed per say. It might be worthwhile to also think about passing an options object in.

rzr:
#923

Issue #377, opened 2016-04-08
Noble + Bleno: Error with noble rejecting unknown commands


popasquat89:
Recently, a change was added to the hci gatt file that rejects commands that are unknown to noble. Anytime that a central goes to discover services on the bleno peripheral, while noble is running on the same device, noble rejects the unknown command. This causes the central that's attempting to discover the services to receive an error, specifically a REQ_NOT_SUPPORTED error.

Commit 6701fc61c5c0f44b948d90f12013ea464baffc80

This is a buzzkill, as we use noble for scanning beacons while at the same time using bleno as a peripheral. Perhaps, we can roll back this commit?

Thanks,
Bryce

sandeepmistry:
> Perhaps, we can roll back this commit?

@popasquat89 rolling back is not likely, however we could introduce a flag to disable this behaviour (maybe an environment variable, that bleno sets).

popasquat89:
I'll nest the conditional logic under the flag NOBLE_MULTI_ROLE, so that any other conditional logic in the future for making the two work nicely together can be supported without new ENV flags.

Bryce

popasquat89:
@sandeepmistry See #379

sandeepmistry:
Closed via #379

PR #379, opened 2016-04-11
multi role - introduce NOBLE_MULTI_ROLE flag for ignoring peripheral role commands

popasquat89:
See #377 

Patch 6701fc61c5c0f44b948d90f12013ea464baffc80 introduced default functionality in noble to send error responses for commands meant for peripheral roles. This is problematic for use with bleno, as all commands that bleno is meant to receive will send back REQ_NOT_SUPPORTED. 

The introduction of the NOBLE_MULTI_ROLE flag will prevent this functionality, as well as provide a flag for all future multi role patches to be placed under. 

Bryce

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,9 +421,17 @@ Example, specify ```hci1```:
 sudo NOBLE_HCI_DEVICE_ID=1 node <your file>.js
 ```
 
+### Bleno compatibility
+
+By default noble will send an error whenever a peripheral command is received. If you're intention is to use bleno in tandem with noble, the following environment variable can be used to bypass this functionality.
+
+```sh
+sudo NOBLE_MULTI_ROLE=1 node <your file>.js
+```
+
 ### Reporting all HCI events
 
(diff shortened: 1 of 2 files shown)


sandeepmistry:
@popasquat89 thanks!

Should bleno auto-set ```NOBLE_MULTI_ROLE``` internally?

popasquat89:
I'm trying to think of any scenarios where you would use both bleno and
noble and not want to enable that flag. The only one I can think of is if
you specified the BLENO_HCI_DEVICE_ID different from the
NOBLE_HCI_DEVICE_ID. Would it make sense in this case for the new flag to
be disabled?

Do we add a flag in bleno to disable the default functionality of enabling
the multi role flag in noble?

I think it might be best to let the developer decide to enable the flag.

Bryce

sandeepmistry:
@popasquat89 thanks!

> I think it might be best to let the developer decide to enable the flag.

Sounds good, let's stick with this for now

PR #380, opened 2016-04-13
Support "long writes" when using hci socket on Linux

projectgus:
Working with some OS X based developers, we noticed the OS X backend automatically performs a "Write Long Characteristics Values" (spec section 4.9.4) operation if you write a characteristic value that's longer than the MTU.

Under the HCI backend on Linux, the characteristic value was truncated "on the wire" and an invalid Bluetooth PDU was sent.

This PR adds automatic "Write Long Characeristics" behaviour on Linux, to match the OS X behaviour (I believe this is spec compliant in both cases, section 4.9.4 doesn't put any restrictions on when it is to be used.)

Please let me know if I need to make stylistic/structure changes, I'm new to node.js.  Thanks for a very useful and powerful project!

### DIFF
--- a/lib/hci-socket/gatt.js
+++ b/lib/hci-socket/gatt.js
@@ -18,6 +18,10 @@ var ATT_OP_READ_BY_GROUP_REQ        = 0x10;
 var ATT_OP_READ_BY_GROUP_RESP       = 0x11;
 var ATT_OP_WRITE_REQ                = 0x12;
 var ATT_OP_WRITE_RESP               = 0x13;
+var ATT_OP_PREPARE_WRITE_REQ        = 0x16;
+var ATT_OP_PREPARE_WRITE_RESP       = 0x17;
+var ATT_OP_EXECUTE_WRITE_REQ        = 0x18;
+var ATT_OP_EXECUTE_WRITE_RESP       = 0x19;
 var ATT_OP_HANDLE_NOTIFY            = 0x1b;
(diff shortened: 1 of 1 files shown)


projectgus:
The only catch I can think of is that there might be a race condition if you perform long writes to multiple characteristics at once without waiting for confirmation that the first write has succeeded. This is because the "execute write request" doesn't specify a characteristic handle, it executes all prepared writes across all handles.

I think it would be possible to work around this by keeping count of the number of longWrite operations currently in progress, and only sending the Execute Write Request when that count goes back to zero - but I don't know enough about hci-socket-js error handling to be sure I can implement this without potentially creating worse bugs.

(FWIW I still think the potential race here is preferable to the old behaviour, where no valid data was sent.)

sandeepmistry:
@projectgus thank you! This change looks good, just need to find a peripheral to test it with locally.

projectgus:
Hi @sandeepmistry,

Awesome. :) I actually just pushed a new version of this change that works around the potential race condition I mentioned in the comment above. And it also happens to be much simpler code, hooray!

Let me know if I can help at all with testing.

Angus

sandeepmistry:
@projectgus thanks! (Also, sorry for the delay)

I've merged and made a few minor changes in https://github.com/sandeepmistry/noble/commit/08a4eb38cd1184c34fba7bc46c0cfc14d3bc5f6d.

Issue #392, opened 2016-04-28
'discover' not seeing manufacturerData on OSX Yosemite


farfromrefug:
Hi,

I am building an few apps around BLE:
- BLE server on rPi using bleno
- iOS / Android native app to connect to that BLE device
- electron app for OSX / Windows using noble

Now everything works perfectly between the rPi and iOS / Android
I have an issue with noble though.
When discovering the devices noble does not report the ```manufacturerData``` on OSX
Though i see it well from And and iOS (which as a very similar code to noble).

on mi Pi i see that i create the adv correctly
```
creating BLEDevice 12334 12wef3
createData: {
  "name": "NeOsmia 12wef3",
  "manufacturerData": [
    3,
    186
  ],
  "services": [
    "6e400001b5a3f393e0a9e50e24dcca9e"
  ]
}
adv [ <Buffer 02 01 06 11 06 9e ca dc 24 0e e5 a9 e0 93 f3 a3 b5 01 00 40 6e>,
  <Buffer 0f 08 4e 65 4f 73 6d 69 61 20 31 32 77 65 66 33 03 ff 03 ba> ]
on -> stateChange: poweredOn true
startAdvertising [ <Buffer 02 01 06 11 06 9e ca dc 24 0e e5 a9 e0 93 f3 a3 b5 01 00 40 6e>,
  <Buffer 0f 08 4e 65 4f 73 6d 69 61 20 31 32 77 65 66 33 03 ff 03 ba> ]
on -> advertisingStart: success
services [ '6e400001b5a3f393e0a9e50e24dcca9e' ]
```
i use
```
                bleno.startAdvertisingWithEIRData(this.advertisement[0], this.advertisement[1]);
```

Now on noble when get the ```discover``` event i get this:
```
parseAdvertisingData { localName: 'raspberrypi',
  txPowerLevel: undefined,
  manufacturerData: undefined,
  serviceData: [],
  serviceUuids: [ '6e400001b5a3f393e0a9e50e24dcca9e' ] }
```
Where manufacturerData is undefined.

Do you have any idea?

Thanks

farfromrefug:
Actually just found something really weird. Using your examples, look what i see
```
peripheral discovered (3dfcac45fee74f3f8176f9e6b409a746 with address <5c:f3:70:68:17:2a, unknown>, connectable true, manufacturerData undefined, RSSI -59:
	hello my local name is:
		raspberrypi
	can I interest you in any of the following advertised services:
[...]

PR #395, opened 2016-05-09
fix missing advertisment on OSX #392

farfromrefug:
As explained in the issue `discover` events on OSX were almost always missing advertisment data

This is actually normal as as gap events are not implemented correctly. You might not get all advertisment data in one event. You need to merge them and trigger when necessary.
It's explained here [](https://learn.adafruit.com/introduction-to-bluetooth-low-energy/gap#advertising-and-scan-response-data)

I use a "hacky" solution because i could not do it the right way because i had no way of getting the scan response data .
So for now if i see that the advertisment data changed i send another `discover` event.

For info it is [done correctly](https://github.com/sandeepmistry/noble/blob/master/lib/hci-socket/gap.js#L192) on linux

BTW wouldn't it be better to use [CBCentralManager](https://developer.apple.com/library/mac/documentation/CoreBluetooth/Reference/CBCentralManager_Class/) on OSX?

### DIFF
--- a/lib/noble.js
+++ b/lib/noble.js
@@ -8,6 +8,41 @@ var Service = require('./service');
 var Characteristic = require('./characteristic');
 var Descriptor = require('./descriptor');
 
+Object.equals = function( x, y ) {
+  if ( x === y ) return true;
+    // if both x and y are null or undefined and exactly the same
+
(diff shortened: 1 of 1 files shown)


yackermann:
@jacobrosenthal Is there a reason this issue  was closed?

PR #399, opened 2016-05-17
fix notify callback to not ignore state.

mikamix:
You are ignoring first arg of notify callback, always passes null.


sandeepmistry:
@mikamix thank you for submitting this.

The first argument of the notify callback is an error arg. #278 adds a seconds argument for the state ....

Issue #305, opened 2016-01-22
Scan response, differences between mac & linux


rosek86:
Hi,

It seems that on Linux Noble tries to merge advertisement and scan response together, which is a good way to handle this. However, if manufacturerData is presence in both; advertisement and scan response Noble returns only scan response field. Issue is with gap.js, line 137:  

https://github.com/sandeepmistry/noble/blob/master/lib/hci-socket/gap.js#L137

I believe that best way is to merge manufacturer data fields together, like it's done on iOS:

      case 0xff: // Manufacturer Specific Data
	if (advertisement.manufacturerData !== undefined) {
          advertisement.manufacturerData = Buffer.concat([advertisement.manufacturerData, bytes]);
  	} else {
          advertisement.manufacturerData = bytes;
        }

The mac version shows different behaviour, It seems that I get two separate events for advertisement and scan response.

BR,
Krzysztof

sandeepmistry:
@rosek86 can you please share an example of the manufacturer data in the advertisement data and scan data, and how CoreBluetooth on iOS and OS X report it?

rosek86:
@sandeepmistry This is something I get on OS X:
Advertisement:  <Buffer 54 44 01 d4 21 54 00 42 03 01 00 00 00 13 00 02 00 00>
Scan response: <Buffer 54 44 01 3b 3d 28 85 35 dd>

There are two separate events, first two bytes is manufacturer identifier which in this case is the same. The iOS merges advertisement and scan response manufacturer data into single field, removing second redundant manufacturer identifier.   

"kCBAdvDataManufacturerData": <544401d4 21540042 03010000 00130002 0000013b 3d288535 dd>]

Then on linux noble returns only scan response data:
<Buffer 54 44 01 3b 3d 28 85 35 dd>

I haven't checked what iOS does if manufacturer identifier is different in advertisement and scan response.

sandeepmistry:
@rosek86 thanks for sharing more details.

> I haven't checked what iOS does if manufacturer identifier is different in advertisement and scan response.

Could you please get more details this? Once we know the behaviour, it's a straightforward change to make ...

rosek86:
@sandeepmistry sorry for late reply with this, this is how this looks with different manufacturer identifiers compared to the same manufacturer identifiers: 

<5444010d ff62005c 03002100 00010000 00 554501 bf73fe1d cfcd> - 5444, then 5545
[...]

PR #423, opened 2016-07-09
Support multiple manufacturer data sections in advertisements

bradjc:
Add support for a manufacturer specific data section in each of the advertisement and the scan response packet. If there are both, they will be returned as an array to the user.

Both fields (if they exist) will also be stored and returned if a future advertisement doesn't have any manufacturer specific fields.
### Notes
1. This is backwards compatible. If there is only one manufacturer section in only some packets it will be preserved and presented with each incoming advertisement.
2. This only supports one manufacturer section in the advertisement and scan response. It is not general for any number of manufacturer sections. The reason is it is difficult to maintain backwards compatibility if someone is using the stored manufacturer data feature.
3. If it is ok to break backwards compatibility, it would be easy to support any number of manufacturer sections by erasing them on each advertisement (like the service uuids and data).

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ peripheral = {
     localName: "<name>",
     txPowerLevel: <int>,
     serviceUuids: ["<service UUID>", ...],
-    manufacturerData: <Buffer>,
+    manufacturerData: <Buffer>, // Can also be an array of buffers if there are multiple sections
     serviceData: [
         {
             uuid: "<service UUID>"
--- a/lib/hci-socket/gap.js
+++ b/lib/hci-socket/gap.js
@@ -94,6 +94,10 @@ Gap.prototype.onHciLeAdvertisingReport = function(status, type, address, address
 
   var discoveryCount = previouslyDiscovered ? this._discoveries[address].count : 0;
   var hasScanResponse = previouslyDiscovered ? this._discoveries[address].hasScanResponse : false;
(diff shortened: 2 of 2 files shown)


sandeepmistry:
Hi @bradjc,

Thank you for submitting this pull request.

Do you have example EIR data that you need to use with this change?

It might be similar to some changes in the [combine-manufacturer-data](https://github.com/sandeepmistry/noble/tree/combine-manufacturer-data) which are also discussed in #305.

I'd like to compare you're sample EIR data, with how iOS reports it as discussed in #305.

bradjc:
I really just need to get manufacturer data from both the scan response and the advertisement. I don't particularly care how. Concatenating them seems weird to me because my use case has distinct messages in each of the manufacturer sections. But I can also see a use case where you stretch data across the two packets because it won't fit in one.

I don't know if this is helpful, but here is the output of `advertisement-discovery.js` with my patch and my device:

```
peripheral discovered (c098e5400019 with address <c0:98:e5:40:00:19, public>, connectable true, RSSI -59:
	hello my local name is:
		sp2101w
	can I interest you in any of the following advertised services:
		[]
	here is my manufacturer data:
		"e002193134312e3231322e31312e3136370000"
		"e0021a616263646566676800"
```
(I did hack the example a bit to display the two manuf. data strings)

jacobrosenthal:
Not against it, but would need multiplatform logs and to get all bindings in line ...

PR #426, opened 2016-07-12
Add raw eir and scanResponse to advertisement object

jslawek:
Hi,
for my project I need access to raw advertisement's eir and scan response hex data. So I added them to advertisement object in gap.js.

### DIFF
--- a/lib/hci-socket/gap.js
+++ b/lib/hci-socket/gap.js
@@ -97,10 +97,12 @@ Gap.prototype.onHciLeAdvertisingReport = function(status, type, address, address
 
   if (type === 0x04) {
     hasScanResponse = true;
+    advertisement.scanResponse=eir;
   } else {
     // reset service data every non-scan response event
     advertisement.serviceData = [];
     advertisement.serviceUuids = [];
+    advertisement.eir=eir;
   }
 
   discoveryCount++;


sandeepmistry:
Hi @jslawek,

Thanks for submitting this. Could you please provide more details as to why your project needs access to the raw EIR data?

While this feature can be implemented in Linux, it will not be possible on Mac OS X.

jslawek:
Hi,
I need to create a very exact "clone" of peripheral device that I do not know before. It needs to be copied with exact (even not compliant with standard) advertisements, services, characteristics, descriptors along with their handle numbers. 
First, using noble, I listen for advertisements (EIR, EIR received as scan response). Next, I scan the peripheral's services, along with handle numbers. I can get the handle numbers or raw descriptors from internal ._noble._bindings._gatts[peripheral.id] objects, so for that I don't need to patch noble. Unfortunatelly, it is not possible with advertisements, as the first "eir" value is overwritten in gap.js as soon as scan response is received. I would be glad to include vanilla noble in my project, that's why I created this PR. I did expect trouble with Mac, but my tool is supposed to work only in Linux.

BTW.  would you be interested in bleno PR described below?
Next, the device description read by noble in first step, is written to editable json. Following that, modified bleno part of the tool reads this json as input and advertises the cloned device. The handle numbers have to exactly match the original device, otherwise the GATT cache of mobile OS will prevent connections to services from smartphone. So, for this part I had to modify bleno to:
- disable automatic set-up of generic services
- enable setting arbitrary handles and descriptors
I noticed you are not a fan of direct operations on handles:
https://github.com/sandeepmistry/noble/pull/290
but for my project this is essential. Maybe it would be useful as an option for others, too?

I will push my tool to Github in about 2 weeks.

cheers

sandeepmistry:
Hi @jslawek, thanks for providing more info. on your project.

> I would be glad to include vanilla noble in my project, that's why I created this PR. I did expect trouble with Mac, but my tool is supposed to work only in Linux.

Yes, understood. However, I'd like to keep functionality between OS X and Linux roughly the same for now. People are also porting the noble API to other platforms, the more Linux specifics we add will result in reduced portability.

For example:
 * @monteslu is working on a Web Bluetooth shim in #238 
 * @jacobrosenthal  is working on a React Native plugin (for iOS and Android): https://github.com/jacobrosenthal/react-native-ble

kaYcee:
Hi @sandeepmistry, I would ask you to reconsider exposing *eir* data in `lib/hci-socket/gap.js`. It would be just one additional field in the `advertisement` object which would allow for custom parsing and delivering/storing the advertisement data in a compact format for later processing.

rzr:

PR #429, opened 2016-07-15
Cant access any uuids you didnt originally 'filter' for

jacobrosenthal:
Getting:
```DOMException: Origin is not allowed to access the service. Tip: Add the service UUID to 'optionalServices' in requestDevice() options. https://goo.gl/HxfxSQ```

Instead of the recommended solution, Ive found it best to simply require users to request all uuids they want to interact with, and then we put each of those in their own services filter.

https://webbluetoothcg.github.io/web-bluetooth/#example-filter-by-services


jacobrosenthal:
oops. wrong repo

PR #430, opened 2016-07-16
little corebluetooth help

jacobrosenthal:
dont merge yet. wip.


lib/mac/bindings.js
```diff
@@ -184,6 +184,69 @@ NobleBindings.prototype.discoverDescriptors = function(identifier, serviceIdenti
...
+  characteristic.once('valueUpdate', function(value, error) {
+    this.emit('read', identifier, serviceIdentifier, characteristicIdentifier, value, null);
+  }.bind(this));
+};
+
+NobleBindings.prototype.write = function(identifier, serviceIdentifier, characteristicIdentifier, data, withoutResponse) {
+  var characteristic = this._characteristicForIdentifier(identifier, serviceIdentifier, characteristicIdentifier);
+
+  // TODO convert type here or in core-bluetooth peripheral?
+  characteristic.writeValue(data, type);
```
jacobrosenthal: Does type need to be converted at all?

lib/mac/bindings.js
```diff
@@ -184,6 +184,69 @@ NobleBindings.prototype.discoverDescriptors = function(identifier, serviceIdenti
   characteristic.discoverDescriptors();
 };
 
+NobleBindings.prototype.read = function(identifier, serviceIdentifier, characteristicIdentifier) {
+  var characteristic = this._characteristicForIdentifier(identifier, serviceIdentifier, characteristicIdentifier);
+
+  characteristic.readValue();
+
+  //can we get isNotification from corebluetooth?
+  characteristic.once('valueUpdate', function(value, error) {
```
jacobrosenthal: can we get isNotification from corebluetooth

lib/mac/bindings.js
```diff
@@ -184,6 +184,69 @@ NobleBindings.prototype.discoverDescriptors = function(identifier, serviceIdenti
...
+    }.bind(this));
+  }
+};
+
+NobleBindings.prototype.readValue = function(identifier, serviceIdentifier, characteristicIdentifier, descriptorIdentifier) {
+  var descriptor = this._descriptorForIdentifier(identifier, serviceIdentifier, characteristicIdentifier, descriptorIdentifier);
+
+  descriptor.readValue();
+
+  // TODO this emits the handle right? but we dont have access to those anymore right?
```
jacobrosenthal: this emits the handle right? but we dont have access to those anymore right?

lib/mac/bindings.js
sandeepmistry: Type will be evaluated as a bool in node-core-bluetooth, so I think it's ok.

lib/mac/bindings.js
sandeepmistry: Unfortunately this was removed from CoreBluetooth at some point.

lib/mac/bindings.js
sandeepmistry: `descriptor.identifier` is the handle, so we can emit if needed.

However, since node-code-bluetooth doesn't support `readHandle` and `writeHandle` I don't think we should.

lib/mac/bindings.js
jacobrosenthal: what do you like to return in that case, null instead of undefined?

sandeepmistry: So the [Yosemite binding](https://github.com/sandeepmistry/noble/blob/master/lib/mac/yosemite.js#L522) is using `false` now, because `args.kCBMsgArgIsNotification` is undefined.  I'm ok with undefined for now we can update the read me accordingly, to say something like true or false if known, otherwise undefined.

jacobrosenthal:
For posterity, heres some testing on timing between corebluetooth and xpc calls. 
```
Jacobs-MacBook-Air:noble jacobrosenthal$ node examples/timetest.js 
xpc bindings
nobleBindings.startScanning object creation: 0.061ms
startScanning roundtrip: 2.388ms
nobleBindings.kCBMsgId37 object creation: 0.966ms
nobleBindings.kCBMsgId37 uuidToAddress: 23.797ms
stopScanning roundtrip: 0.324ms

^C
Jacobs-MacBook-Air:noble jacobrosenthal$ node examples/timetest.js 
corebluetooth bindings
CentralManager.prototype.scanForPeripherals object creation: 3.594ms
startScanning roundtrip: 6.225ms
centralManager:didDiscoverPeripheral object creation: 20.583ms
stopScanning roundtrip: 0.779ms
^C
```

At best it looks like its about ~100us extra per roundtrip call for an argument-less call like stopScan. But were seeing a 1-2 ms per object creation, most notably in discovering a peripheral where it adds up to ~20ms. Theres also a 'missing' ~1ms in the start scan function, presumably due to passing the argument.

However the hack to lookup uuidToAddress was previously taking 23ms and no one seemed to mind so the transition seems doable.

PR #431, opened 2016-07-17
Tests

jacobrosenthal:
Pretty minimal as these things go. Its been a while..  What do you think.

### DIFF
--- a/lib/mac/bindings.js
+++ b/lib/mac/bindings.js
@@ -287,7 +287,7 @@ NobleBindings.prototype._onPeripheralDiscover = function(peripheral, advertiseme
     localName: advertisementData.localName,
     txPowerLevel: advertisementData.txPowerLevel,
     manufacturerData: advertisementData.manufacturerData,
-    serviceData: advertisementData.serviceData,
+    serviceData: advertisementData.serviceData ? advertisementData.serviceData : [],
     serviceUuids: convertUuids(advertisementData.serviceUuids)
   };
 
@@ -331,11 +331,9 @@ function convertUuid(uuid) {
 }
 
 function convertUuids(uuids) {
-  var convertedUuids;
+  var convertedUuids = [];
 
   if (uuids) {
-    convertedUuids = [];
-
     uuids.forEach(function(uuid) {
       convertedUuids.push(convertUuid(uuid));
     });
--- a/lib/mac/bindings.js
+++ b/lib/mac/bindings.js
@@ -240,8 +240,12 @@ NobleBindings.prototype.notify = function(identifier, serviceIdentifier, charact
   }.bind(this));
 }
 
-NobleBindings.prototype.init = function() {
-  this._centralManager = new CoreBluetooth.CentralManager();
+NobleBindings.prototype.init = function(centralManager) {
+  if(centralManager) {
+    this._centralManager = centralManager;
+  }else {
+    this._centralManager = new CoreBluetooth.CentralManager();
+  }
 
   this._centralManager.on('address', this._onAddress.bind(this));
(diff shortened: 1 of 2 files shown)


lib/mac/bindings.js
```diff
@@ -283,7 +287,7 @@ NobleBindings.prototype._onPeripheralDiscover = function(peripheral, advertiseme
     localName: advertisementData.localName,
     txPowerLevel: advertisementData.txPowerLevel,
     manufacturerData: advertisementData.manufacturerData,
-    serviceData: advertisementData.serviceData,
+    serviceData: advertisementData.serviceData ? advertisementData.serviceData : [],
```
sandeepmistry: There's some inconsistencies yosemite.js and node-core-bluetooth here:
 * https://github.com/sandeepmistry/node-core-bluetooth/blob/master/lib/central-manager-delegate.js#L68
 * https://github.com/sandeepmistry/noble/blob/master/lib/mac/yosemite.js#L140

One is an object, the other is an array, but it should be an array. Unrelated to this PR, but we need to correct node-core-bluetooth.

Then we don't need this check?

sandeepmistry:
I think `test-bindings.js` needs to be renamed to something like `test-core-bluetooth-bindings.js` or moved to a `mac` folder, so other binding tests can be added later.

Other than that looks good.

jacobrosenthal:
Agreed. Id actually like to figure out a way to abstract these tests out for all bindings layers and make them a requirement somehow. I asked @phated for a good example of this and he mentioned LevelUp/LevelDown's abstract https://github.com/Level/abstract-leveldown

jacobrosenthal:
Theres really 2 sets of tests going on in here, one is testing the bindings interface to make sure that startScanning emits scanStart, etc. Another test is testing the native specific layer, so that if the native layer emits a peripheralDiscover event its properly transformed into a discover event. Maybe these should be pulled apart and the bindings interface shared

jacobrosenthal:
I still havent though of how exactly to implement the tests into the abstraction. In all the tests Id ideally like the bindings author to not even have control or access to the test or at least the final assertion. They should simply set up some internal state, and stub any functions they need to to make the proper hardcoded assertion fall out..

phated:
https://github.com/Level/memdown/blob/master/test.js#L8-L30

jacobrosenthal:
Yeah they dont do any stubbing or anything though. Feel free to show me

jacobrosenthal:
Aight Converted test-bindings-central  Not so sure about it yet.

test/test-bindings-characteristic.js
```diff
@@ -0,0 +1,79 @@
...
+
+  it('should emit read', function() {
+    var eventSpy = sandbox.spy();
+    bindings.once('read', eventSpy);
+
+    //stub native
+    sandbox.stub(mock.nativeCharacteristicObject, "readValue", function(){
+      this.emit('valueUpdate', a.dataBuffer);
+    });
+
```
sandeepmistry: Were you thinking we could create test helpers?

For example, tests for another bindings might have a very similar setup to this, so maybe there's an opportunity to share some test code between them?

You could pass in the binding, maybe have hook to setup stubs and another for verify ...

sandeepmistry:
I've started some tests for the HCI socket binding in the [`more-test`](https://github.com/sandeepmistry/noble/tree/more-test) branch based on the infrastructure @jacobrosenthal has created in the this PR for those that are interested.

jacobrosenthal:
converted the rest of the tests and brought them into react-native-ble. So far pattern seems solid. 

Ive asked @monteslu to report any problems he finds with web ble. @sandeepmistry wanna keep going and see if anything else comes up with hci?

rzr:
#923

PR #432, opened 2016-07-22
Update hci.js

vdharmon:
Added this._deviceId property

### DIFF
--- a/lib/hci-socket/hci.js
+++ b/lib/hci-socket/hci.js
@@ -70,6 +70,7 @@ var Hci = function() {
   this._socket = new BluetoothHciSocket();
   this._isDevUp = null;
   this._state = null;
+  this._deviceId = null;
 
   this._handleBuffers = {};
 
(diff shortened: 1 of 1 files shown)


lib/hci-socket/hci.js
```diff
@@ -86,7 +87,7 @@ Hci.prototype.init = function() {
 
   var deviceId = process.env.NOBLE_HCI_DEVICE_ID ? parseInt(process.env.NOBLE_HCI_DEVICE_ID, 10) : undefined;
 
-  this._socket.bindRaw(deviceId);
+  this._deviceId = this._socket.bindRaw(deviceId);
```
sandeepmistry: Hi @vdharmon,

From my understanding, `bindRaw` does not return anything, should this line be `this._deviceId = deviceId;`?

vdharmon: Since this._socket is actually a BluetoothHciSocket object, when I follow the code through to the BluetoothHciSocket class I'm seeing it's bindRaw method return an int BluetoothHciSocket::bindRaw(int* devId) { .. } (I have BluetoothHciSocket 5.0 installed).  Would that not be correct?  When I write out the this._deviceId value  I always get what was defined with NOBLE_HCI_DEVICE_ID parameter which is set through the bindRaw call.

sandeepmistry: @vdharmon yes agreed. I was looking at the usb.js instead of the C++ code in `BluetoothHciSocket.cpp`. Sorry for the confusion.

PR #438, opened 2016-07-29
Added extraction of the solicitation service UUID fields

smartyw:
As discussed, modified gap.js to extract the three types of solicitation service UUID fields.


sandeepmistry:
@smartyw thank you!

PR #452, opened 2016-08-28
Support for long write (writeHandle; write without response)

jslawek:
Hi,
for my project I use writeHandle. I noticed it does not support long writes. So I have added the longWriteHandle (basically copied longWrite with minor changes) in gatt.js, and modified the writeHandle. 

BTW. I think long write should also be possible with standard writes without response. So I have also modified the gatt write to support this.

### DIFF
--- a/lib/hci-socket/gatt.js
+++ b/lib/hci-socket/gatt.js
@@ -525,12 +525,12 @@ Gatt.prototype.read = function(serviceUuid, characteristicUuid) {
 Gatt.prototype.write = function(serviceUuid, characteristicUuid, data, withoutResponse) {
   var characteristic = this._characteristics[serviceUuid][characteristicUuid];
 
-  if (withoutResponse) {
+  if (data.length + 3 > this._mtu) {
+    return this.longWrite(serviceUuid, characteristicUuid, data, withoutResponse);
+  } else if (withoutResponse) {
     this._queueCommand(this.writeRequest(characteristic.valueHandle, data, true), null, function() {
       this.emit('write', this._address, serviceUuid, characteristicUuid);
     }.bind(this));
-  } else if (data.length + 3 > this._mtu) {
-    return this.longWrite(serviceUuid, characteristicUuid, data, withoutResponse);
   } else {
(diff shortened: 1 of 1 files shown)


peturh:
What do you mean by long write? That you can send larger packets then... what? I've been struggling with this in my project for quit some time.

jslawek:
A single write command is limited in length (max packet length - MTU, depending on parameters negotiated during connection initialization). That is why the Bluetooth core specification defined so-called "long write", which allows you to write longer requests, by splitting them in chunks. It consists of several requests: first, the "prepare write" for consecutive chunks, then "execute write". See section 4.9.4 "Write Long Characteristic Values" of core spec pdf.
Noble has support for long writes already included (check out the longWrite prototype in lib/hci-socket/gatt.js):

https://github.com/sandeepmistry/noble/blob/a05e7f87bd10447ec397070a39d33c86768641fc/lib/hci-socket/gatt.js#L545-#L585

However, at this moment it is called only if you invoked write with response:

https://github.com/sandeepmistry/noble/blob/a05e7f87bd10447ec397070a39d33c86768641fc/lib/hci-socket/gatt.js#L528-#L533

I think it would be beneficial for noble users to call the long write also in standard writes (one part of my PR). So, until it is patched, if you use noble "write" and the long write does not work for you, you can try to write with response.

If you by any chance use writeHandle (like me), I have also included a separate prototype function for long writing with handles. I believe Sandeep will be able to come out with something smarter, e.g. wrap the existing longWrite which operates on  characteristics and services UUIDs.

peturh:
Thank you, this makes a lot of sense to me. I'm talking over a project from another developer and he has implemented his own queue algorithm on top of noble using `withoutResponse = true`, chunking up the packet to 20 bytes a piece then sending it with `peripheral.writeHandle(...)`.

sandeepmistry:
Hi @jslawek,

Thank you for submitting this pull request.

I don't think writes without responses should use long writes, so if you remove this behaviour I am open to merging.

Also is it possible to re-use the new `longWriteHandle` function inside of `longWrite` they look very similar.

sandeepmistry:
Closing for now due to lack of activity ...

Issue #471, opened 2016-10-04
Multiple discover & connection


Lahorde:
I'm able to do multiple discover successfully. But if I do a discover -> Connection -> Disconnection -> discover. In last discover advertising devices are not found. I do not have any LE Advertising reports 

After checking hcidump and code I saw scan parameters are only set in hci.js l502 : 
```
    } else if (this._state !== 'poweredOn') {
      this.setScanEnabled(false, true);
      this.setScanParameters();
    }
```
So in my case, scan parameters are only set before first scanning. 
But after looking bluetooth datasheet I did not found any info about scan parameters persistence in HCI. So maybe it is better to always send a Set Scan Parameters before scanning.

 Here is modified gap.js , I always set scan parameters :

```
Gap.prototype.startScanning = function(allowDuplicates) {
  this._scanState = 'starting';
  this._hci.setScanEnabled(false, !allowDuplicates);
  this._hci.setScanParameters();
  this._hci.setScanEnabled(true, !allowDuplicates);
};
```
Here is code to test : 

```
var async = require('async');

var NobleDevice = require('../index');

var TestDevice = function(peripheral) {
    NobleDevice.call(this, peripheral);
};

NobleDevice.Util.inherits(TestDevice, NobleDevice);

TestDevice.discover(function(testDevice) {
    console.log('found ' + testDevice.id);

    testDevice.on('disconnect', function() {
        console.log('disconnected!');
        //process.exit(0);
    });

    async.series([
        function(callback) {
            console.log('connect');
            testDevice.connect(callback);
        },
        function(callback) {
            console.log('discoverServicesAndCharacteristics');
            testDevice.discoverServicesAndCharacteristics(callback);
        },
        function(callback) {
            console.log('readDeviceName');
            testDevice.readDeviceName(function(error, deviceName) {
                console.log('\tdevice name = ' + deviceName);
                callback();
            });
        },
        function(callback) {
            console.log('disconnect');
            testDevice.disconnect(callback);
        },
        function(callback) {
            console.log('discover');
            TestDevice.discover(function(testDevice){
                console.log('found ' + testDevice.id);
                callback();
            });
        }
    ]
    );
});
```

Here are hci logs, when no LE advertising reports are get in second advertising

```
< HCI Command: Set Event Mask (0x03|0x0001) plen 8
    Mask: 0xfffffbff07f8bf3d
> HCI Event: Command Complete (0x0e) plen 4
[...]

PR #473, opened 2016-10-06
Issue https://github.com/sandeepmistry/noble/issues/471

Lahorde:
Always set scan parameters before an active scanning.

Now scan parameters are only set when an hci device becomes ready : 
refer Hci.prototype.pollIsDevUp and this code that is called when device becomes up :

 in hci.js l550 :
```
    if (hciVer < 0x06) {
      this.emit('stateChange', 'unsupported');
    } else if (this._state !== 'poweredOn') {
      this.setScanEnabled(false, true);
      this.setScanParameters();
    }
```
I would remove :

      this.setScanEnabled(false, true);
      this.setScanParameters();

and only set scan parameters when a scan is needed. 
But in code scan parameters response is used in device up processus, when it is received a :

    this.emit('stateChange', 'poweredOn');

is sent.I would suggest emitting this event when all responses command sent in  pollIsDevUp have been received. But we can also sending this event in :

 in hci.js l550 :
```
    if (hciVer < 0x06) {
      this.emit('stateChange', 'unsupported');
    } else if (this._state !== 'poweredOn') {
      //this.setScanEnabled(false, true);
      //this.setScanParameters();
      this.emit('stateChange', 'poweredOn');
    }
```

### DIFF
--- a/lib/hci-socket/gap.js
+++ b/lib/hci-socket/gap.js
@@ -23,7 +23,12 @@ util.inherits(Gap, events.EventEmitter);
 Gap.prototype.startScanning = function(allowDuplicates) {
   this._scanState = 'starting';
   this._scanFilterDuplicates = !allowDuplicates;
-
+  
+  // Always set scan parameters before scanning 
+  // https://www.bluetooth.org/docman/handlers/downloaddoc.ashx?doc_id=229737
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@Lahorde thanks!

PR #483, opened 2016-10-18
Fix some typos in documentation!

bhaveshgohel:


sandeepmistry:
Hi @bhaveshgohel,

Thanks for taking the time to submit this PR. I've purposely styled it this way to indicate the parameter(s) were optional. Is the way you've changed it to a more standard way of doing this?

bhaveshgohel:
Pardon me, I didn't correctly understand : "Is the way you've changed it to a more standard way of doing this?"


The main intention of the PR was that other person can simply copy paste the code and use it by simply removing callback with there functions. I should have also remove the "["  "]" in that case.

With the comma it looks like typo. But I totally get your point of purposely styling to indicate it as optional. It was just a suggestion from my side. :)

sandeepmistry:
For example, looking at the Node.js Buffer docs: https://nodejs.org/dist/latest-v4.x/docs/api/buffer.html#buffer_new_buffer_str_encoding

They use the style:
> new Buffer(str[, encoding])

To indicate `encoding` is an optional parameter.

Let's stick with the current approach, as it's a similar style to what Node.js is use for the official docs. I understand your concern about copying and pasting, thanks for clarifying.

Issue #295, opened 2016-01-03
Crypto error when connecting to WICED Sense


deadprogram:
This is to replace https://github.com/sandeepmistry/node-bluetooth-hci-socket/issues/15

When connecting to the Broadcom WICED Sense tag from Linux (Ubuntu 14.04), I hit the following error. I do not have this problem with any other BLE hardware:

```
buffer.js:229
      length += list[i].length;
                       ^

TypeError: Cannot read property 'length' of undefined
    at Function.Buffer.concat (buffer.js:229:24)
    at Object.c1 (/home/ron/Development/cylonjs/cylon-example/node_modules/cylon-ble/node_modules/noble/lib/hci-socket/crypto.js:8:19)
    at Smp.handlePairingResponse (/home/ron/Development/cylonjs/cylon-example/node_modules/cylon-ble/node_modules/noble/lib/hci-socket/smp.js:86:12)
    at Smp.onAclStreamData (/home/ron/Development/cylonjs/cylon-example/node_modules/cylon-ble/node_modules/noble/lib/hci-socket/smp.js:57:10)
    at emitTwo (events.js:92:20)
    at emit (events.js:172:7)
    at AclStream.push (/home/ron/Development/cylonjs/cylon-example/node_modules/cylon-ble/node_modules/noble/lib/hci-socket/acl-stream.js:35:10)
    at NobleBindings.onAclDataPkt (/home/ron/Development/cylonjs/cylon-example/node_modules/cylon-ble/node_modules/noble/lib/hci-socket/bindings.js:252:15)
    at emitThree (events.js:97:13)
    at emit (events.js:175:7)
```

sandeepmistry:
@deadprogram as mentioned in https://github.com/sandeepmistry/node-bluetooth-hci-socket/issues/15   "As well as include what you've done so far to debug the issue ..."

It's really difficult for me to support you with out more info from you.

deadprogram:
That error was all I had at that point. Let me locate my WICED Sense and see if I can get additional details. Any suggestions on how to output some useful debug info?

sandeepmistry:
@deadprogram HCI dump would be useful.

As well as you spending sometime to better understand what was going wrong, etc.

deadprogram:
Hi, @sandeepmistry I have some further details. Note that this device was working with the versions of Noble from before the hci-socket refactor.

[...]

PR #495, opened 2016-11-07
Register onAclStreamDataBinded after sendPairingRequest is called

BasKiers:
Changed hci-socket smp to only register onAclStreamDataBinded after sendPairingRequest is called. This fixes a crash where the 'data' event called onAclStreamDataBinded before sendPairingRequest was called as mentioned in issue #295

### DIFF
--- a/lib/hci-socket/gatt.js
+++ b/lib/hci-socket/gatt.js
@@ -150,7 +150,7 @@ Gatt.prototype.onAclStreamData = function(cid, data) {
 };
 
 Gatt.prototype.onAclStreamEncrypt = function(encrypt) {
-  if (encrypt) {
+  if (encrypt && this._currentCommand) {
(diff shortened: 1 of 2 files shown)


lib/hci-socket/gatt.js
```diff
@@ -150,7 +150,7 @@ Gatt.prototype.onAclStreamData = function(cid, data) {
 };
 
 Gatt.prototype.onAclStreamEncrypt = function(encrypt) {
-  if (encrypt) {
+  if (encrypt && this._currentCommand) {
```
sandeepmistry: Can you please provide an HCI dump for when this case is triggered? (and description as well).

The changes in `lib/hci-socket/smp.js` look good to me!

maxdenisov:
Wiced sense is working with this fix now.

PR #500, opened 2016-11-14
Added 'Adapter specific known issues' section to the README. 

gotleg:
First entry is how to fix Error: Command disallowed with some BLE adapters when trying to connect.


Ashishjshetty:
Hi @gotleg,
I think the right place for this might be [noble-wiki](https://github.com/sandeepmistry/noble/wiki) instead, maybe create a known issues section and add it there, this can later be updated with any other future discoveries.

sandeepmistry:
@gotleg thanks!

@Ashishjshetty the wiki suggestion is a good idea, if this section in the read me gets to big, we can definitely move it over to the wiki and link to it in the read me.

PR #505, opened 2016-11-22
notifications and read callbacks Fixes #490

### LINKED ISSUE #490: notifications and read callbacks
I'm trying to use the callback from characteristic.read(), but my device also has notifications. i'm subscribed to these, so i also have a characteristic.on('data') handler. 

but things get mixed up, the callback of the read() function sometimes returns the data from the notification, not the actual data which i was reading - though, I get that value too in the 'data' event handler.

I think this is a bug, I have some workarounds, but they cause other unfortunate limitations :(

zkiiito:
I'm not sure this is still backward compatible, or not, or what should it be backward compatible with.


sandeepmistry:
@zkiiito thanks!

PR #514, opened 2016-12-05
[DO NOT MERGE] Bump connection parameters

sandeepmistry:
Can be tested via `npm install sandeepmistry/noble#connection-interval-test`


tybro0103:
So these connection interval settings helped my connection issues on RPI significantly. But what's with the timeout? It went from `200` (`0x00c8 `) to `42`. That part made things much worse. For my use case a timeout closer to 1 second is ideal. Have you considered exposing these settings as part of the API?

jacobrosenthal:
Might be taken care of in a 2.0 dbus change.

Issue #276, opened 2015-12-02
Error handling?


rclai:
I notice that every action called (writes in particular as well as others) assumes a successful outcome as seen by how the callback is always called with a null error object.

Is this a bluetooth rule of thumb that errors should never be emitted back?

sandeepmistry:
@rclai these are placeholders for now, errors still need to be bubbled up for some actions.

See https://github.com/sandeepmistry/noble/issues/184 and https://github.com/sandeepmistry/noble/issues/207.

tybro0103:
+1

tybro0103:
https://github.com/sandeepmistry/noble/pull/564

Bonitis:
@tybro0103 this may not be the best place for this; do you have any best practices for handling errors with noble? I'm dealing with a lot of errors in connections, service discovery, and characteristic discovery. This can usually be resolved by trying to re-connect to the device. Would it cause problems or be overly expensive to call `noble.startScanning()` whenever I hit an error in the connection/discovery process?

PR #536, opened 2017-01-20
Do report errors in characteristics discovery

MarSoft:
Before this change, `peripheral.discoverSomeServicesAndCharacteristics`
silently ignored any discovery errors, as a result any error led to callback not being called.

### DIFF
--- a/lib/peripheral.js
+++ b/lib/peripheral.js
@@ -79,13 +79,33 @@ Peripheral.prototype.discoverServices = function(uuids, callback) {
 
 Peripheral.prototype.discoverSomeServicesAndCharacteristics = function(serviceUuids, characteristicsUuids, callback) {
   this.discoverServices(serviceUuids, function(err, services) {
+    if(err) {
+      if(callback) {
+        callback(err);
+      }
+      return;
+    }
+    
     var numDiscovered = 0;
     var allCharacteristics = [];
+    var failed = false;
 
     for (var i in services) {
       var service = services[i];
 
       service.discoverCharacteristics(characteristicsUuids, function(error, characteristics) {
+        if(failed) {
+          // some other characteristic failed and we passed error to the caller, so do nothing anymore
+          return;
+        }
+        if(error) {
+          failed = true;
+          if(callback) {
+            callback(error);
+          }
+          return;
+        }
+        
         numDiscovered++;
 
         if (error === null) {


MarSoft:
Just found that this change itself is pointless, as both `discoverServices` and `discoverCharacteristics` are made such that they never report errors themselves...
Could anybody take a look and think of some way to fix that?

tybro0103:
😱😱😱

Just discovered the same. And it isn't just the two methods you mentioned; it seems to be the pattern throughout the whole library. I noticed it when tracking down an issue unsubscribing from a characteristic. All these callbacks are documented with an `error` param, even though the library makes no attempt to ever pass one.

So I analyzed it a bit...

1. [`peripheral.discoverServices` calls `noble.discoverServices`](https://github.com/sandeepmistry/noble/blob/master/lib/peripheral.js#L77)
2. [`noble.discoverServices` in turn calls `bindings.discoverServices`](https://github.com/sandeepmistry/noble/blob/master/lib/noble.js#L188)
3. [`bindings.discoverServices` will emit a `servicesDiscover` event](https://github.com/sandeepmistry/noble/blob/master/lib/webbluetooth/bindings.js#L188)
4. [noble listens for that event and invokes `noble.onServicesDiscover()`](https://github.com/sandeepmistry/noble/blob/master/lib/noble.js#L30)
5. [`noble.onServicesDiscover` emits `servicesDiscover` on `peripheral`](https://github.com/sandeepmistry/noble/blob/master/lib/noble.js#L30)
6. [`peripheral.discoverServices` from #1 listens for that event to invoke the callback](https://github.com/sandeepmistry/noble/blob/master/lib/peripheral.js#L72)

The problem is that errors are going to happen within the bindings, and it looks as though bindings just aren't expected to send errors back up the chain. Look at the [method signature of `noble.onServicesDiscover`](https://github.com/sandeepmistry/noble/blob/master/lib/noble.js#L191)...no error param! The WebBluetooth bindings are even [catching a number of errors](https://github.com/sandeepmistry/noble/blob/master/lib/webbluetooth/bindings.js#L352), but don't have a way to pass them back up the chain.

On another note, there's a lot of delegation going on from `peripheral`/`characteristic`/etc to `noble` to `bindings` (and back), and it seems like it'd be a lot safer for them to be using callbacks instead of events.

tybro0103:
Looks like it's been brought up a few times before:
https://github.com/sandeepmistry/noble/issues/276

tybro0103:
An idea for fixing this: https://github.com/sandeepmistry/noble/pull/564

rzr:

PR #541, opened 2017-01-27
Fix destructive query behavior for services, characteristics and desc

Chondrules:
This pull request changes the default behavior of the noble library to be non destructive when issuing queries to devices.  With this patch the services, characteristics and descriptors get tested for existance before being overwritten with a fresh empty object, if they already exist they remain and the current array is always returned.

I tested this against the CC2650 sensortag and am now able to maintain multiple open data streams simultaneously with the device while also doing other device queries without halting already existing notifications.


lucascoupez:
👍

PR #557, opened 2017-02-22
Simpler install instructions for Windows

don:
use windows-build-tools instead of Visual Studio

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,14 +62,19 @@ Make sure you have read and write permissions on the ```/dev/usb/*``` device tha
 
 ### Windows
 
- * [node-gyp requirements for Windows](https://github.com/TooTallNate/node-gyp#installation)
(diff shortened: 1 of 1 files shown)


OwenBrotherwood:
I would advice that one waits with this as there are perceived problems until full enlightenment is obtained

https://github.com/nodejs/node-gyp/issues/1116

Current recommendation from my opinionated view with requirement for confirmation within a couple of days is ensure that one follows MS very closely with http://landinghub.visualstudio.com/visual-cpp-build-tools and the doc on the page and avoid more dependency hell: MS and c++ the necessary evils, other modules are not. 
(off topic ref Linus and his many rants about c contra c++: dependency hell should be avoided)

sandeepmistry:
@don thanks!

Issue #363, opened 2016-03-27
Persistent services and characteristics across disconnect and reconnect


mattgol:
First of all thanks for noble - it's the most mature piece of BLE software out there!  Noble made it very simple to connect to my homemade peripherals and made my project possible in the first place - but now I'm facing an issue regarding battery life: I plan to connect to the device every couple of minutes, read a characteristic, and disconnect. Unfortunately, I have to go through the service discovery process every time I reconnect - which increases the number of necessary bluetooth packages and this means less battery life.

Is there a way to store the GATT structure across a disconnect/reconnect cycle? This would solve my issue.

Right now there are two workarounds I can use:
1. Keep the connection but with long connection interval and slave latency (this can be requested by the slave). This allows me about 30 sec in between packets but not longer as the hardware doesn't accept anything longer as supervision timeout.
2. Use readHandle/writeHandle to read/write directly. This means I have to change the server script every time the GATT structure changes. Also this wouldn't work on OSX but this doesn't hurt me on my linux box.

Anybody out there who had the same issue?

ayhoung:
Have you tried storing the peripheral object in cache and reconnecting to it after a while?

ayhoung:
@mattgol could you share how you are using readHandle/writeHandle to read/write directly?

mattgol:
@ayhoung I dug into the source code of noble-device and noble and also ran everything in the node console interactively to have a look into the objects. I also had btmon running to look at the actual communication with the BT hardware. I could trace the issues down to the internal representation in the moble master object.This is my current status (everything is on linux and looks different in OSX):
- The peripheral object is constructed in the library during the discovery and returned to the user. After service discovery it contains the structure of the services and characteristics. 
- The noble master object contains an array of the handles to the open connections to translate the object id to the handle of the connection. The handle number matches the output of 'hcitool con'. This list is updated after the reconnect so this is not the issue.
- The noble master object also keeps track of the GATT handles. The read method of a characterics object then calls the read method in the noble master object which then looks up the GATT handle number in its list. The actual communication with the BT hardware does not care about the characteristics structure - it only uses the GATT handle numbers. This is by design for BLE communication as it cuts down the transmitted bits to a minimum this way. 
- The noble master object then catches the event with the returned data and relays the data to the user (via the callback). 
[...]

PR #559, opened 2017-02-23
Preserve GATT structure across connections

mattgol:
Running on Linux the GATT structure is currently not preserved after a disconnect. On OSX it's possible to scan for services, disconnect, reconnect and directly access characteristics without scanning for services again - simply by reusing the service and characteristic objects received from an earlier scan. The current version of noble unfortunately removes all internal references to services, characteristics, and descriptors for Linux after disconnect: After each connection the services have to be scanned again which costs time and battery life of the device. The code in this PR keeps the objects around and simply restores the references after a reconnect. (See also [#363](https://github.com/sandeepmistry/noble/issues/363).) 

The discussion happened quite some time ago but recently I had the opportunity to also test on OSX. The behavior is consistent between operating systems and compliant to the Bluetooth standard as far as possible without major changes to noble:
Typically the GATT structure is very stable and therefore the Bluetooth standard encourages reusing the GATT structure across connections. If the structure changes at any time the device has to notify the communication partner by an indication sent from the service changed characteristic (service 1801, characteristic 2a05). If this characteristic does not exist the structure is guaranteed to stay put. 

OSX does everything under the hood: We don't get access to the service 1800 and 1801 at all from noble or any other application. This probably means that OSX itself monitors for changed services. Therefore a completely consistent behavior between OSes would require a major rewrite for Linux which possibly breaks all applications that read service 1800 to get names etc. 

The next best thing is to leave everything to the application level: Any user of noble on Linux can subscribe to characteristic 2a05 and perform service discovery afterwards. (As in most cases the structure changes only with firmware upgrades an application would also get away without monitoring at all. But this would not be standards compliant.) 

This change helped a lot to increase battery life for my devices: I use a long running application to read environment data every 10 mins and I already get months of battery life out a coin cell.

### DIFF
--- a/lib/hci-socket/bindings.js
+++ b/lib/hci-socket/bindings.js
@@ -5,6 +5,7 @@ var util = require('util');
 
 var AclStream = require('./acl-stream');
 var Gatt = require('./gatt');
+var GattStructure = require('./gatt-structure');
 var Gap = require('./gap');
 var Hci = require('./hci');
 var Signaling = require('./signaling');
@@ -22,6 +23,7 @@ var NobleBindings = function() {
 
   this._handles = {};
   this._gatts = {};
+  this._gattStructures = {};
   this._aclStreams = {};
   this._signalings = {};
 
@@ -184,9 +186,12 @@ NobleBindings.prototype.onLeConnComplete = function(status, handle, role, addres
     uuid = address.split(':').join('').toLowerCase();
 
     var aclStream = new AclStream(this._hci, handle, this._hci.addressType, this._hci.address, addressType, address);
-    var gatt = new Gatt(address, aclStream);
-    var signaling = new Signaling(handle, aclStream);
 
+    if(!this._gattStructures.hasOwnProperty(uuid)) {
+      this._gattStructures[uuid] = new GattStructure();
+    }
+    var gatt = new Gatt(address, aclStream, this._gattStructures[uuid]);
+    var signaling = new Signaling(handle, aclStream);
     this._gatts[uuid] = this._gatts[handle] = gatt;
     this._signalings[uuid] = this._signalings[handle] = signaling;
     this._aclStreams[handle] = aclStream;
--- a/lib/hci-socket/gatt-structure.js
+++ b/lib/hci-socket/gatt-structure.js
@@ -1,0 +1,12 @@
+//
+// Gatt Structure Encapsulation
+//
+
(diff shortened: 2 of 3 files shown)


ccamrobertson:
@mattgol what version of OSX are you seeing the GATT structure being preserved across connections? (with or without your PR?)

mattgol:
Hi @ccamrobertson ! At the time I wrote this I was still on Yosemite - I didn't want to upgrade as my Mac refused to boot after the upgrade to El Capitan. In the meantime I did a complete reinstall to Sierra. Now my Mac shows the same problem: After a disconnect/connect cycle I have to do a rescan. Unfortunately, my PR only targets the linux/windows side - sorry.

ccamrobertson:
@mattgol thanks for confirming. I believe that the issue is in El Capitan; I adding debugging to the bindings and saw nothing come through in this case. Given that this change hasn't been merged in yet I'm wary of opening a general issue on the main project, but will keep an eye out if others see the same thing.

mattgol:
@ccamrobertson 
The behavior is different compared to Linux. On stock noble under Linux the characteristic handles are removed from noble after connection which this PR fixes. On El Capitan/Sierra the characteristic can be called an actually emits an event after reading but it only returns data on the first connection and simply returns undefined after reconnect. I don't know how to fix this though - the bug is in the native part which I don't know enough about.

ccamrobertson:
@mattgol while I didn't continue testing on OS X, I can verify that this preserves the GATT structure across requests, as intended, on my C.H.I.P. - nice work!

@sandeepmistry any chance this can be tee'd up for review and merge?

jrobeson:
@sandeepmistry : do you like this approach? If so, then it'd be nice to have.

wind-rider:
@sandeepmistry I'm also interested to hear the outcome from this!

yackermann:
High Sierra same issue

rzr:
Please also check with:

mattgol:
@rzr: Thanks for sorting out the changes from abandonware! Since there was no activity here for a while I also based my current fork on abandonware's version. A lot of the open PRs and a lot of other useful changes did go into the their fork but this issue remained open.
Therefore, I've added the changes mentioned in the PR (and some minor other changes) for my own usage. I didn't do a PR there as I understand now that my changes are not perfect for everyone - the Gatt structure is never deleted which is basically a memory leak whenever someone constantly connects different devices.

PR #572, opened 2017-03-16
More descriptive error outputs

mbifulco:
As helpful as it is to have the Status Mapper print out plain english error codes, It might also be nice to give users the HCI error code they're dealing with.  I found [this table](http://www.lisha.ufsc.br/teaching/shi/ine5346-2003-1/work/bluetooth/hci_commands.html) which seems to mirror what's returned in that status field - so in this PR, I convert `status` to a hex string and spit it out with the plaintext.

### DIFF
--- a/lib/hci-socket/bindings.js
+++ b/lib/hci-socket/bindings.js
@@ -214,8 +214,10 @@ NobleBindings.prototype.onLeConnComplete = function(status, handle, role, addres
     this._gatts[handle].exchangeMtu(256);
   } else {
     uuid = this._pendingConnectionUuid;
-    var errorCode = status.toString(16);
-    error = new Error('HCI Error: (0x' + errorCode +  ') ' + Hci.STATUS_MAPPER[status] || ('HCI Error: Unknown (' + status + ')'));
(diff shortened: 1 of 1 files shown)


lib/hci-socket/bindings.js
```diff
@@ -214,8 +214,8 @@ NobleBindings.prototype.onLeConnComplete = function(status, handle, role, addres
     this._gatts[handle].exchangeMtu(256);
   } else {
     uuid = this._pendingConnectionUuid;
-
-    error = new Error(Hci.STATUS_MAPPER[status] || ('Unknown (' + status + ')'));
+    var errorCode = status.toString(16);
+    error = new Error('HCI Error: (0x' + errorCode +  ') ' + Hci.STATUS_MAPPER[status] || ('HCI Error: Unknown (' + status + ')'));
```
sandeepmistry: Two questions:

* should we put the hex code after the string?
* should we also use hex code for unknown errors?

mbifulco: Both changes make sense - I'll amend my PR.

mbifulco:
This should be better now, @sandeepmistry

sandeepmistry:
@mbifulco thanks!

Issue #299, opened 2016-01-09
process hanging?


AKST:
Is there any reason why nobel would cause a program to hang, even after closing `noble.stopScanning()` has been called and no peripherals are connected?

sandeepmistry:
@AKST noble continuously monitoring for adapter state change events: https://github.com/sandeepmistry/noble#adapter-state-change

AKST:
is there a means to stop this once I no longer wish to connect any BLE devices?

sandeepmistry:
@AKST not currently. What is the goal for stopping this?

Is using ```process.exit(0)``` to exit you app an option?

AKST:
Well I’m using this library as part of another library, and I don’t think it’s fair to expect the users of my own library to have to call process.exit(0) to close there applications which is why I’m seeking a means to stop this once the library is no longer being used.

sandeepmistry:
@AKST I've renamed this issue, and labelled it as an enhancement request.

AKST:
Thanks @sandeepmistry, that's fair enough.

I would be happy to help out where I can to progress this issue, so if you feel like delegating any work just let me know, as I went poking through quite a bit of the source to try and find a solution myself before posting this.

Otherwise I'll be more than happy to just test any changes.

sandeepmistry:
If you could help out that would be great!

I think we'll have to add equivalent API's in noble's dependencies, as they continuously listen:
 * https://github.com/sandeepmistry/node-xpc-connection - for OS X
 * https://github.com/sandeepmistry/node-bluetooth-hci-socket - for non-OS X

AKST:
Awesome, I start having a look around the weekend!

— Angus

recursify:
@sandeepmistry Can we also add an  API to *start* noble?

For example, this simple script will hang:

```
require('noble')
```

IMO - requiring a module *should not* hang.  So something like:

```
var noble = require('noble');
noble.start();
// Do stuff ..
noble.stop();
```

sandeepmistry:
@recursify this would be a breaking change, something to start discussing in #325.

sandeepmistry:
@AKST any progress on this?

AKST:
Eh unfortunately not, I probably shouldn't have put my name forward to eagerly. Feel free to delegate it someone else

jayalfredprufrock:
[...]

PR #577, opened 2017-03-20
noble.stop() function

oyooyo:
Lousy first try at implementing a noble.stop() function, as requested in #299 
Probably doesn't even work correctly; I just fiddled around until the test program:
```
noble = require('noble');
noble.stop()
```
seemed to somehow work. Might still be useful as a kind of starting point for a proper stop() function.

### DIFF
--- a/lib/noble.js
+++ b/lib/noble.js
@@ -52,6 +52,10 @@ function Noble(bindings) {
 
 util.inherits(Noble, events.EventEmitter);
 
+Noble.prototype.stop = function() {
+  this._bindings._hci.stop();
+};
+
 Noble.prototype.onStateChange = function(state) {
   debug('stateChange ' + state);
 
--- a/lib/hci-socket/hci.js
+++ b/lib/hci-socket/hci.js
@@ -105,6 +105,11 @@ Hci.prototype.init = function() {
   }
 };
 
+Hci.prototype.stop = function() {
+    this._socket.stop();
+    clearTimeout(this._pollIsDevUpTimeout);
+}
+
 Hci.prototype.pollIsDevUp = function() {
   var isDevUp = this._socket.isDevUp();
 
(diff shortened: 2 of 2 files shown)


oyooyo:
Thanks for the hint @mbifulco , I tried to fix the bug you mentioned.
Not quite sure if I did it correctly though, all this GitHub stuff is still new and confusing to me.

mbifulco:
Totally understand @oyooyo - it can be daunting.  I was able to tell what was wrong before by clicking "details" next to the [travis-ci](https://travis-ci.org/) automation that's running on PRs in this repo - and then navigating into an individual build to see what error was being reported.

pasted_image_3_30_17__4_33_pm

It looks like your builds are passing now, which puts this PR one step closer to being done with.  I'll leave it to @sandeepmistry to comment on whether or not the modification to the code is the best way to stop noble from doing its thing.

lib/noble.js
```diff
@@ -52,6 +52,10 @@ function Noble(bindings) {
 
 util.inherits(Noble, events.EventEmitter);
 
+Noble.prototype.stop = function() {
+  this._bindings._hci.stop();
```
sandeepmistry: It would be nice for the bindings to have their own `.stop()` function instead of calling the one on `_hci`.

Then someone can fill in the mac version and other bindings `stop()` function. For mac we'll need for this to be implemented in a dependency of noble: https://github.com/sandeepmistry/node-xpc-connection/issues/16

jacobrosenthal:
It seems like this should be implicit in the library rather than called by the user api. Can we count active listeners and close bindings if it goes to 0?
would need mac bindings to merge

readeral:
Is this pull request still progressing? Anything else that needs doing to get it merged?

rzr:
#923

PR #585, opened 2017-03-30
linux hack for gatttool command invocation

claymodel:
work around for the 
non standard BLE CCCD (not read able CCCD)

### DIFF
--- a/lib/hci-socket/bindings.js
+++ b/lib/hci-socket/bindings.js
@@ -392,7 +392,10 @@ NobleBindings.prototype.notify = function(peripheralUuid, serviceUuid, character
   var handle = this._handles[peripheralUuid];
   var gatt = this._gatts[handle];
 
-  if (gatt) {
+  if (handle == 0x000f) {
+    gatt.notifySpecial(serviceUuid, characteristicUuid, notify);
+    return;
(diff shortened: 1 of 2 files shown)


lib/hci-socket/bindings.js
```diff
@@ -392,7 +392,10 @@ NobleBindings.prototype.notify = function(peripheralUuid, serviceUuid, character
   var handle = this._handles[peripheralUuid];
   var gatt = this._gatts[handle];
 
-  if (gatt) {
+  if (handle == 0x000f) {
```
sandeepmistry: I'm not too keen on merging this, this change is super specific to your device, other devices could use handle 0x0f and not have this issue.

sandeepmistry:
@claymodel thanks for submitting this, unfortunately based on my comment above I'm not keen on merging this.

What's the issue you are facing? What BLE peripheral has the issue? I'd much prefer to fix the underlying issue instead of the device specific workaround you proposed.

jrobeson:
I think it'd be fair to close this and let them reopen it if they want to elaborate.

sandeepmistry:
@jrobeson good suggestion!

@claymodel please see @jrobeson's comment above.

PR #593, opened 2017-04-09
add stream() to characteristic

jacobrosenthal:
To start the discussion.

Ive been working with streams on the serialport beta. One conclusion I think Im coming to is I dont think . streams lifecycle should have anything to do with connection lifecycle. You dont want a stream closing your bluetooth connection and having to reopen if you want to start another stream after.

Hence You have to get and keep a connection outside of the stream. Im putting it on the characteristic, and you can generate a new stream() repeatedly. 

I think letting from2 through2 and duplexify handle stream setup and teardown is the current thinking, and certainly way easier to implement so Im using those libraries rather than trying to subclass Readable Writable or readable-stream

I think a stream is pointless if you're not subscribed, and you'll lose data inbetween subscribe and unsubscribes, so I think you should have to have previously subscribed to characteristic, or characteristic.stream() will attempt to do that for you.

This means Im holding the subscribed state on the characteristic object now.

usage would look something like this

```
    from2([Buffer.alloc([0x00, 0x00,0x00])])
      .pipe(characteristic.stream())
      .pipe(to2.obj(function (data, enc, cb) {
        console.log(data);
        this.end(); // flush writable
        cb(); //callback writable
        }, process.exit)
      );
  });
```

A more advanced example, but Im finding I want to send and receive commands, the characteristic stream will just keep asking for buffers from the from2 before weve gotten responses back. Im fixing that like this

```
    var sourcer = from2();
    sourcer.push(cmdList.shift());

    sourcer
      .pipe(characteristic.stream())
      .pipe(nmgr.decode())
      .pipe(utility.hashToStringTransform())
      .pipe(to2.obj(function (data, enc, cb) {
        console.log(data);
        var cmd = cmdList.shift()
        if(!cmd){
          sourcer.push(null); //close readable
          this.end(); // flush writable
        }else{
          sourcer.push(cmd);
        }
        cb(); //callback writable
        }, process.exit)
      );
```
It looks much better as a pull stream, but thats a different discussion, and characteristic.stream() can be converted to pull stream successfully without leaking so implementing as 'standard' node 2/3 stream is the way to go


jacobrosenthal:
Sandeep pointed me at Max's work from a while ago. 
https://github.com/maxogden/ble-stream/blob/master/central.js

He attempts to wrap lifecycle up into creation, and also polls read, which I dont like.

@sandeepmistry do we need this write throttling?
https://github.com/maxogden/ble-stream/blob/master/central.js#L43

sandeepmistry:
> do we need this write throttling? https://github.com/maxogden/ble-stream/blob/master/central.js#L43

Really depends on the connection interval and buffer depth. I can see data being dropped if too much is pushed. We'll have to try it out ...

lib/characteristic.js
```diff
@@ -15,6 +18,7 @@ function Characteristic(noble, peripheralId, serviceUuid, uuid, properties) {
   this.type = null;
   this.properties = properties;
   this.descriptors = null;
+  this.subscribed = 0;
```
sandeepmistry: nit pick, is this false/true or 0/1?

jacobrosenthal: bindings are 0/1 and later i assign what they return

sandeepmistry: Ok, I was expecting it to be true/false in the bindings.

jacobrosenthal:
I still want this :) 

biggest reason not to merge this is adding like <15 deps
    "end-of-stream": "^1.0.0",
    "inherits": "^2.0.1",
    "core-util-is": "~1.0.0",
    "inherits": "~2.0.3",
    "isarray": "~1.0.0",
    "process-nextick-args": "~1.0.6",
    "safe-buffer": "~5.1.1",
    "string_decoder": "~1.0.3",
    "util-deprecate": "~1.0.1"
    "stream-shift": "^1.0.0"

what were all these deps attempting to solve outside of core?
https://r.va.gg/2014/06/why-i-dont-use-nodes-core-stream-module.html
looks like trying to standardize between all the .8 .9 .10 streams streams2 streams3 problems

what node are we supporting in noble?  "node": ">=0.8", so wed still want to use these deps presumably

so assuming we 2.0 noble at some point, are streams ready for primetime in core?

working group seems to track here
https://github.com/nodejs/readable-stream/issues

Doesnt look like anything big added in last few releases.. node 9 uses new node errors in streams
node 8 stream now inherit destroy method
node 7 bug fixes?
node 6 null is now an invalid chunk to write in object mode

Guess wed have to implement to see whats up

jacobrosenthal:
Yeah definitely,  Frankly though, the more I research and understand the conclusion Im coming to is that streams are a dead experiment nobody is working on and are afraid to touch.

I dont really understand it all yet, but async iterators are coming with tc39, and observables are also a proposal. These will be in v8 shortly and are going to be the cross platform solution thats hooked up to async await and promises the way streams will never be.

rzr:
Please forward the change to:

PR #597, opened 2017-04-17
Allow duplicate uuid's for services, characteristics, and descriptors.

### LINKED ISSUE #175: peripherals with duplicate services or characteristics are not handled properly

brycejacobs:
This patch is initial work to fix #175 

The main idea is to replace our UUID lookups with ids that are generated locally, so that if a service, characteristic, or descriptor shares a UUID, noble won't overwrite the service repeatedly.

I've only started with the HCI bindings, and will turn my eye to all of the others once @sandeepmistry gives me the nod that this method will satisfy all concerns of his.

I think the PR is simple enough, but let me know if there are any concerns.

### DIFF
--- a/lib/mac/legacy.js
+++ b/lib/mac/legacy.js
@@ -12,6 +12,7 @@ var uuidToAddress = require('./uuid-to-address');
 
 var XpcConnection = require('xpc-connection');
 
+var serviceNum = 0;
 var NobleBindings = function() {
   this._peripherals = {};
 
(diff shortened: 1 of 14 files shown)


brycejacobs:
Ping @sandeepmistry 

If this is the general direction you were talking about by adding an ID, let me know and I'll translate this across the remainder of the bindings.

sandeepmistry:
There's some work in https://github.com/sandeepmistry/noble/commit/996a1bdc96bcd435fb82a0b7365f572dfbec4737 you can base off, it's close to what you have.

I think the service id's can just be the start handle of the service, the characteristic id's can be the characteristic handle etc.. What do you think?

cc/ @jacobrosenthal

brycejacobs:
@sandeepmistry That sounds good, if you look at my OSX code I actually started heading in that direction once I realized the handle made more sense.

Food for thought...what if we made the UUID a symbol, instead of depending on the handle id for it's unique property. It makes us not have to store the handle, and makes each UUID unique by nature of the Symbol. For backwards compatibility, something like [es6-symbol](https://www.npmjs.com/package/es6-symbol) should work fine.

So we could have

var services = {
  (new Symbol('13333333333333333333333333333337')): { ... }
  (new Symbol('13333333333333333333333333333337')): { ... }
}

Each of the services have to be accessed by the instantiated symbol. This would somewhat replicate what OSX does natively when using CoreBluetooth. You receive an instantiated UUID, rather than a string representation of the UUID, so one key doesn't match another key if the UUID is the same.

I think this method makes more sense than making the user pass the serviceId, characteristicId, or descriptorId to the surface API calls.

brycejacobs:
Let me know your thoughts @sandeepmistry and @jacobrosenthal , and I'll try to roll the patch tomorrow while I have some time dedicated.

jacobrosenthal:
Sorry for silence.. Been thinking about this. Not sure what is gained from going to an es6-symbol object vs string uuid. You like it because we can still store the same uuid in two different symbol objects and they will still be distinct?

brycejacobs:
@jacobrosenthal Correct. This would kinda resemble how it is when you program in let's say iOS, you receive an instantiated UUID object that contains the UUID string but will be distinct from a separate service with a redundant UUID.

```javascript
(new Symbol('13333333333333333333333333333337')) === (new Symbol('13333333333333333333333333333337')) // false

(new Symbol('13333333333333333333333333333337')).toString() === (new Symbol('13333333333333333333333333333337')).toString() // true

// We could also go with the typical class.
function UUID(uuid, id) { this.uuid = uuid; }

const idOne = new UUID('1);
const idTwo = new UUID('1');

idOne === idTwo // false
```
The only reason that I didn't recommend with the class way, is because I thought it was cleaner to have the symbol be passed to the binding. Also, we would need to instantiate the UUID class from within the binding. Seems cleaner to use a language feature here instead.

I.E. from within Service.js
```javascript
this._noble.discoverCharacteristics(peripheralId, symbolUUID, characteristicUUIDs)

vs.

this._noble.discoverCharacteristics(peripheralId, classUUID, characteristicUUIDs)
```
The only reason it makes sense in the first place to make it unique. If I need to lookup a service by a uuid, a simple string won't help me since I could have multiple services with the same uuid.

I'm good with either, or simply leaving the API to where it has to pass the service handle in. Just let me know what you prefer.

brycejacobs:
@sandeepmistry @jacobrosenthal I just went ahead and used the id. Please let me know if there is anything you would like me to fix with the PR.

Thanks guys!

brycejacobs:
@sandeepmistry @jacobrosenthal Any input?

lib/mac/legacy.js
```diff
@@ -12,6 +12,7 @@ var uuidToAddress = require('./uuid-to-address');
 
 var XpcConnection = require('xpc-connection');
 
+var serviceNum = 0;
```
sandeepmistry: Is this a left over?

brycejacobs:
Appreciate the look, let me answer as best I can.

1.) I wasn't  sure if this was a WIP, so I left this alone. Let me look again and see what I can do.

2.) I can take a look and do a check on the arguments so we can discern how it's being used. What exactly are you looking to achieve out of this though?

The application I'm using this on is a hub, which acts as a multi role system which has slaves of the same type. These slaves are all represented with the same service. I'll try to come up with a simple example using bleno if I can. (Not sure if bleno allows the same uuid for different services)

sandeepmistry:
> 1.) I wasn't  sure if this was a WIP, so I left this alone. Let me look again and see what I can do.

Thanks

> 2.) I can take a look and do a check on the arguments so we can discern how it's being used. What exactly are you looking to achieve out of this though?

Something like:

```javascript
function Service(noble, peripheralId, id, uuid) {
  if (arguments.length === 3) {
    // old args were: noble, peripheralId, uuid
    uuid = id;
  }

  // ...
}
```

jacobrosenthal:
Related to #316, I think were going to look at refactoring ids in 2.0 instead of rewriting 1.0

PR #599, opened 2017-04-18
Don't initialize bindings until first state change listener added

jacobrosenthal:
Align with bleno
https://github.com/sandeepmistry/bleno/pull/293/files

### DIFF
--- a/lib/noble.js
+++ b/lib/noble.js
@@ -9,6 +9,7 @@ var Characteristic = require('./characteristic');
 var Descriptor = require('./descriptor');
 
 function Noble(bindings) {
+  this.initialized = false;
   this.state = 'unknown';
   this.address = 'unknown';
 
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
Tested on web bluetooth bindings and this indeed fixes a state change issue I was bugging @monteslu about.  osx bindings worked as always. Lazy init also cleans up my logs so I dont see ble init noise when Im conditionally including but not using noble. Need to investigate react-native-ble possibilities.

jacobrosenthal:
Now that I use this more, its problematic actually.

For the common pattern of adding a statechange listener, its perfect, but many times people dont wait for statechange and just call scan (for instance on web ble, by the time a person clicks the button, ble is certainly available) -- except its not because im not attaching a listener to wait for it.

So it either has to be backed out ...

OR the only actual noble api call is actually startscanning, so we can check that bindings havent been initted yet, and do a once('statechange') to init them.

Im going to prototype that today.

jacobrosenthal:
Scratch that, theres a bunch of api calls wed have to check in. 

I think we have to either back this out, or call the startscan without waiting onStateChange an antipattern and see how many of those wacky folks are out there...

jacobrosenthal:
Continuing discussion in #606

PR #606, opened 2017-04-25
more work for lazy bindings

jacobrosenthal:
We merged the last change, but Im already seeing some oddness with folks who use the startscanning without waiting for (or even creating a listener on) statechange thus not having bindings init. Thinking about it more, someone might also be polling noble.state and would never init bindings either.

The only other public facing api is noble.stopScan but thats easier, if were not init, just dont call the bindings layer.

### DIFF
--- a/lib/noble.js
+++ b/lib/noble.js
@@ -10,9 +10,9 @@ var Descriptor = require('./descriptor');
 
 function Noble(bindings) {
   this.initialized = false;
-  this.state = 'unknown';
-  this.address = 'unknown';
 
+  this.address = 'unknown';
+  this._state = 'unknown';
   this._bindings = bindings;
   this._peripherals = {};
   this._services = {};
@@ -48,12 +48,27 @@ function Noble(bindings) {
     }
   }.bind(this));
 
+  //lazy init bindings on first new listener, should be on stateChange
   this.on('newListener', function(event) {
     if (event === 'stateChange' && !this.initialized) {
       this._bindings.init();
       this.initialized = true;
     }
   }.bind(this));
+
+  //or lazy init bindings if someone attempts to get state first
+  Object.defineProperties(this, {
+    state: {
+      get: function () {
+        if (!this.initialized) {
+          this._bindings.init();
+          this.initialized = true;
+        }
+        return this._state;
+      }
+    }
+  });
+
 }
 
 util.inherits(Noble, events.EventEmitter);
@@ -61,7 +76,7 @@ util.inherits(Noble, events.EventEmitter);
 Noble.prototype.onStateChange = function(state) {
   debug('stateChange ' + state);
 
-  this.state = state;
+  this._state = state;
 
   this.emit('stateChange', state);
 };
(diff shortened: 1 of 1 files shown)


lib/noble.js
```diff
@@ -48,20 +48,35 @@ function Noble(bindings) {
...
+  //lazy init bindings on first new listener, should be on stateChange
   this.on('newListener', function(event) {
     if (event === 'stateChange' && !this.initialized) {
       this._bindings.init();
       this.initialized = true;
     }
   }.bind(this));
+
+  //or lazy init bindings if someone attempts to get state first
+  Object.defineProperties(this, {
```
sandeepmistry: Overall this PR looks good to me.

Any idea what versions of Node.js added support for `Object.defineProperties`?

jacobrosenthal:
Looks like ecma2015.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties

Never used this site before, but says node 5
http://node.green/

the MDN has a javascript polyfill posted...

sandeepmistry:
Ok, I think it would be a breaking change if we add this in without a polypill. The Travis CI currently does smoke tests as far back a v0.8.0 of Node.

How do you think we should proceed?

jacobrosenthal:
Are tests currently failing? I presume they would be. Ill add the polyfill
and see if tests pass?

sandeepmistry:
I think we're lacking coverage, so everything is still passing.

Maybe we need a smoke test to just require('noble') for now?

jacobrosenthal:
Well solve that later :) Blaine said its been in all version of node.

0.8.x had unrelated install problems on my osx. But I tested in 0.10.x on my machine

```
^CJacobs-MacBook-Air:noble jacobrosenthal$ node -v
v0.10.48
Jacobs-MacBook-Air:noble jacobrosenthal$ git branch
* delay-init
  master
  newtmgr
  streams
  test
  webble-disconnect
Jacobs-MacBook-Air:noble jacobrosenthal$ node examples/advertisement-discovery.js 
peripheral discovered (b09f6cd52341407cbb9da1db6ff1d0a4 with address <unknown, unknown>, connectable true, RSSI -87:
	hello my local name is:
		undefined
	can I interest you in any of the following advertised services:
		[]
	here is my manufacturer data:
		"4c000c0e00a38404b2e2e7e9f7afd71f206a"

peripheral discovered (8b551de822f74909a7d2dcb18065a30a with address <unknown, unknown>, connectable true, RSSI -80:
	hello my local name is:
		undefined
	can I interest you in any of the following advertised services:
		[]
	here is my manufacturer data:
		"4c000215014d6b984df5b53831e55d26947591f500010001c8"

^CJacobs-MacBook-Air:noble jacobrosenthal$ 
```
Lets get a merge in?

sandeepmistry:
Cool, yes please go ahead and merge. I won't be able to test for a few days ...

sandeepmistry:
@jacobrosenthal could you please port this to bleno as well?

jacobrosenthal:
Roger.

Issue #625, opened 2017-05-31
128-bit service data advertisement is not parsed.


arekzelechowski:
OS: Linux ARM v7
BlueZ v5

Currently if we look at [gap.js](https://github.com/sandeepmistry/noble/blob/master/lib/hci-socket/gap.js#L193), we can see that noble parses advertisement/scan response packets to fill `serviceData` property.

According to the [docs](https://www.bluetooth.com/specifications/assigned-numbers/generic-access-profile), packets can specify 16-bit (`0x16`), 32-bit (`0x20`) and 128-bit (`0x21`) UUIDs service data. However, noble does use only 16-bit UUID.

On the other hand, noble parses 128-bit service list (`0x06` and `0x07`).

Example:
`02010604160f18401321cf6ea2ffad809e8d36443b5ab56f6d2b0040` is parsed to:
`{"serviceData":[{"uuid":"180f","data":"5d"}]`
Instead of:
`{"serviceData":[{"uuid":"180f","data":"5d"},{"uuid":"2b6d6fb55a3b44368d9e80adffa26ecf","data":"0040"}]`

What do you think about this?
BR

sandeepmistry:
Closed via #626. Thanks @arekzelechowski!

rzr:
Relate-to: https://github.com/abandonware/noble/issues/81#issuecomment-621761981

PR #626, opened 2017-05-31
Support 32-bit and 128-bit service data UUID (sandeepmistry/noble#625)

arekzelechowski:
Take a look on #625 for more info..


jrobeson:
does this need a test?

sandeepmistry:
@jrobeson I think it's ok like this. We are still working on unit tests with @jacobrosenthal on a few branches.

PR #651, opened 2017-07-17
Binding options

bitcloud:
To use two different devices in one setup you need to configure the binding in your code and not with ENV-Vars.

I am using one device for connecting while the other keeps on collecting advertisements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,6 +467,20 @@ Example, specify ```hci1```:
 sudo NOBLE_HCI_DEVICE_ID=1 node <your file>.js
 ```
 
+If you are using multiple HCI devices in one setup you can run two instances of noble with different binding configurations by initializing them seperatly in code:
+
+```
+const HCIBindings = require('noble/lib/hci-socket/bindings');
+const Noble = require('noble/lib/noble');
+
+const params = {
+  deviceId: 0,
+  userChannel: true
+};
+
+const noble = new Noble(new HCIBindings(params));
+```
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -467,6 +467,20 @@ Example, specify ```hci1```:
...
 
+If you are using multiple HCI devices in one setup you can run two instances of noble with different binding configurations by initializing them seperatly in code:
+
+```
+const HCIBindings = require('noble/lib/hci-socket/bindings');
+const Noble = require('noble/lib/noble');
+
+const params = {
+  deviceId: 0,
+  userChannel: true
```
ghost, line 626: is this a good name for this? I don't really have any suggestions atm, but the name should describe what it does. Especially if it can work across other bindings.

jrobeson:
Is it possible to make this work on OSX? I don't know anything about CoreBluetooth.

bitcloud:
I think device selection was and will always be target platform specific. This is why I suggested to split the bindings in separate packages and added the possibility to pass parameters to those bindings directly. 

I don't think that we have something like userChannel in OSX or even IOs.

I have no idea about coreCoretooth as well, sorry, can't help there.

README.md
bitcloud: This seems to be the name of the source. So even in the kernel driver it is named like that so I thought I stick with that.

rzr:
Please also check with:

PR #652, opened 2017-07-17
added connection canceling to hci lib

bitcloud:
add the ability to cancel connection attempts.

P.S.:
I am using the hci-socket lib directly thats why I just added it there. Wouldn't it be nice to split out the bindings so you can use them directly if you want to go a bit more low level but not the whole way through ;-)


rzr:
Please also check with:

PR #653, opened 2017-07-17
added 5.0 error messages

bitcloud:
added some missing error messages from BT5


sandeepmistry:
@bitcloud thanks! Do you have a reference section of the core spec for the new error codes? Have you experienced these error messages?

bitcloud:
You'll find them in the spec (e.g.): http://www.mouser.li/pdfdocs/bluetooth-Core-v50.pdf
I think I experienced one of the last two. But that was quite a while ago while I was debugging some Bluetooth stuff.

rzr:
May I suggest to submit your patches to:
https://github.com/abandonware/noble

PR #691, opened 2017-09-28
Bump the minimum Node requirement to 4.x

jrobeson:
4.x is now supported across all major distros and their LTS versions (if they have one)

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -2,6 +2,7 @@
   "browser": true,
   "node": true,
   "mocha": true,
+  "esversion": 6,
   "unused": false, //enable after cleanup
   "undef": false, //enable after cleanup
   "globals": {
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,10 @@ os:
   - osx
 language: cpp
 env:
-  - NODE_VERSION="0.8"
-  - NODE_VERSION="0.10"
-  - NODE_VERSION="0.12"
   - NODE_VERSION="4"
   - NODE_VERSION="6"
   - NODE_VERSION="8"
+
 matrix:
   fast_finish: true
 before_install:
@@ -18,9 +16,6 @@ before_install:
   - nvm use --delete-prefix $NODE_VERSION;
   - if [ $TRAVIS_OS_NAME == "linux" ]; then
       export CC="g++-4.8";
-    fi
-  - if [ $NODE_VERSION == "0.8" ]; then
-      npm install -g npm@2;
     fi
 install:
   - npm install
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   ],
   "main": "./index.js",
   "engines": {
-    "node": ">=0.8"
+    "node": ">=4.0"
   },
   "os": [
     "darwin",


jrobeson:
You hadn't decided if you wanted to accept this, but I figured I'd submit it anyways.

jacobrosenthal:
Does the engines comment save node 0.8 users from semver updating them to a build they cant run? Would this need a major version change? Either way, why not go to 6 which is actually the LTS at this point.

jrobeson:
I'd love to, but  in https://github.com/noble/noble/issues/741 you mentioned being concerned about new enough dbus on raspbian and that's likely to be less of a problem than new enough node unless you recommend folks installing from buster-backports.

However you also did say about going into beta for a year. https://github.com/noble/noble/issues/741#issuecomment-362011294 

So in that case, we could jump directly to 8.x which I'd prefer, but a year is a long time without a stable release.  

It might be best to release 2.0 with promises and not break the API, and then jump to node 8.x for 3.x and that can go into beta.

jrobeson:
There are different schools of thought on whether bumping the required langauge version counts as a semver break though.

jacobrosenthal:
Oh I dont mean doing any the stuff in the 2.0 thread at this point, I mean just making the node 6.x bump a 2.0 .. and if @sandeepmistry  is interested in promises throw that in too. Makes the version bump more obvious too.

ghost:
@jacobrosenthal : there doesn't seem to be a reason to bump to 6.x vs 8.x since no current version of debian stable uses 6.x. It's either 4.x now or 8.x for the next release. There's also no LTS version of ubuntu with 6.x either.

jacobrosenthal:
To repeat then, if we move we SHOULD go straight to 8.x?

ghost:
Please don't ask me that :) I'm too biased. I have zero interest myself in supporting debian stable or old ubuntu LTSs. I'm just trying to compromise for everyone elses's sake.

ghost:
The one thing I don't want to do however is to let people who are selling products based on these old ubuntu or debian versions dictate the future of a library they don't even contribute to. I don't know enough about the noble user base to know who is doing that or if it's mostly hobbyists.

ghost:
Let me spam a bit more here.  I've been willing to put a decent amount of time into this library, but I have no idea how to move the process forward.

don:
Going to 4.x is good for now. I wouldn’t worry about having a strong SLA for old versions. We can drop 4.x or 6.x as necessary as things start breaking.

ghost:
it's not about things breaking for me. it's about being able to use block scoped variables, default parameters, rest parameters, spread syntax, destructuring, classes, and async/await. 

As per https://node.green/

ghost:
@don : can you explain what you mean by "as things start breaking".  I could break 4.x support easily with a PR for let and const.

ghost:
Is there any way we can get together and talk about our goals for the library? Get a roadmap going?

jacobrosenthal:
> I've been willing to put a decent amount of time into this library, but I have no idea how to move the process forward.

Going off topic for a sec. And obviously not speaking as any gatekeeper, but:

In my estimation, the real way to help sandeep would be to triage all the issues down now and in the future. Ideally we would start to promote a community (maybe somewhere else?) so the newbies and startup bros could solve their own issues and we could do the fun part of maintaining.

So not to call you out  @jrobeson as youve been present and accounted for more than most, but you're confusing what you want (to help with fun syntactic sugar) with what the library needs (to help with miserable tech support on 115 open issues)

ghost:
bumping to node 4 isn't what I want. I want 8.x.  (and at the time 6.x). I was compromising my needs with those of the community who can't use newer nodes. I feel like i made that clear in being concerned for raspbian/debian stable users? Was it not?

I just looked over the issues, again, and most of them are related to:

 * hardware I don't own: mac or a pi
 * Oses i don't run, like Windows, OSX, old versions of debian
 * bugs in the various bindings like CoreBluetooth or hci-socket (which I don't know much about)
 * bugs or missing features, like errors not being propagated up the stack, 
 * requiring knowing more about the BLE stack than I do.

I can't triage what I don't know much about :(

I'm here mostly as a code janitor.

ghost:
More and more bits of the ecosystem are requiring 6.x now, so I've decided against pursuing 4.x any further.

PR #708, opened 2017-11-02
Support for ES6 Promises

MaxDesiatov:
In this pull request all user-facing APIs that take callbacks now also return promises. This change is backwards compatible. 

Some tests had to be changed, as previously there were expectations that callbacks were called immediately in the same event loop cycle. This is no longer true as callbacks are scheduled with promises, but shouldn't break any user code, which should have been written with an expectation that callbacks are asynchronous in the first place.

Additional tests were added to verify that all promises resolve successfully.

This fork is also published as https://www.npmjs.com/package/noble-promise in the meantime, as I see that main repository had no merging activity for some time. I still hope that this PR will be merged eventually.

@sandeepmistry could you please have a look at this PR? Thanks a lot for the library and I look forward to your feedback.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,5 +1,6 @@
 {
   "browser": true,
+  "esversion": 6,
   "node": true,
   "mocha": true,
   "unused": false, //enable after cleanup
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ os:
   - osx
 language: cpp
 env:
-  - NODE_VERSION="0.8"
-  - NODE_VERSION="0.10"
-  - NODE_VERSION="0.12"
   - NODE_VERSION="4.0"
   - NODE_VERSION="4.1"
   - NODE_VERSION="4.2"
--- a/lib/characteristic.js
+++ b/lib/characteristic.js
@@ -35,8 +35,8 @@ Characteristic.prototype.toString = function() {
 };
 
 Characteristic.prototype.read = function(callback) {
-  if (callback) {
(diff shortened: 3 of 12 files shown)


jrobeson:
so what would it take to get this merged?

jrobeson:
as a review, I'd like to see the actual 4.x upgrade done in a separate PR rather than hidden in this PR.

maxdesiatov:
happy to apply any changes to merge this as required by current maintainers, which I guess are @sandeepmistry and @jacobrosenthal

maxdesiatov:
ok, I see there are a few conflicts. Happy to fix those if there is any indication that is the only thing blocking this from being merge. Otherwise, would love to know what else might be needed

jrobeson:
@maxdesiatov : the Node 4.x changes are here https://github.com/noble/noble/pull/691, so when the time for rebase comes you'll need to remove your node 4.x changes.

maxdesiatov:
no problem, can do that as soon as #691 is merged

jrobeson:
@maxdesiatov : When the goahead is given, then it'd also need some examples.

jacobrosenthal:
Touches a lot of code sadly (mostly because of whitespace) And sad we also have to touch our tests so much. Are people still doing callback AND promise apis? That would let all the existing tests pass to start with and give much more confidence.

jrobeson:
I think the nice part about doing both promises and callbacks is that folks can start using promises immediately and then we could deprecate the callbacks for 3.x

rzr:
#923

Issue #724, opened 2017-12-21
descriptors got from discoverDescriptors() only contain 16-bit uuid even for custom descriptors


sfyang:
Hello,

While writing a script, running on Ubuntu Linux 16.04 with BlueZ 5.37, for finding all the descriptors of a custom characteristic, the uuid of the descriptor objects passed to the callback by discoverDescriptors() of the characteristic contain only 16-bit value. However, since the characteristic contains at least one custom descriptors, the 16-bit value can not be used to  reliably identify the custom descriptor by matching the full 128-bit UUID. Judging from the value I got, I think currently the truncated 16-bit value of the custom descriptor is only the less significant 2-byte of the full 16-byte value.

I did a little code digging, and, since I was running the script on Linux which uses the "hci-socket" binding, it looks like that the problem is caused by the way the "Find Info Response" was parsed between line 668 and 677 in  gatt.js. Is there way to get the full 128-bit UUID for custom descriptors?

AlexPeitz:
Yes you are right. That is because every time a find info request using characteristic handles are sent, the peripheral replies in batches with either 16 bit or 128 bit descriptor UUIDs, with a header containing as first byte the opcode and as second byte an identifier saying if the following payload contains handle-uuid pairs of 16 bit uuids or 128 bit uuids. The line `num = data[1]` reads that identifier, but does not take into account if it is a 128 bit value.

I fixed this by replacing the for loop in line 671 : 

```
for (i = 0; i < num; i++) {
        descriptors.push({
          handle: data.readUInt16LE(2 + i * 4 + 0),
          uuid: data.readUInt16LE(2 + i * 4 + 2).toString(16)
        });
}
```

with the following code:

```
var format = data[1];
      switch(format){
        case 1:
          var num = (data.length - 2) / 4;
          for (var i = 0; i < num; i++) {
            descriptors.push({
              handle: data.readUInt16LE(2 + i * 4 + 0),
              uuid: data.readUInt16LE(2 + i * 4 + 2).toString(16)
            });
          }
        break;
        case 2:
          var num = (data.length - 2) / 18;
          data = data.slice(2);
          for (var i = 0; i < num; i++) {
            var h = data.readUInt16LE(i * 18);
            var uuidBuf = data.slice(2 + i*18, (i+1)*18);
            var u = "";
            for (var j = 0; j < uuidBuf.length; j++) {
              var newChar = uuidBuf.readUInt8(j).toString(16);
              if (newChar.length < 2)
[...]

PR #726, opened 2017-12-23
Add files via upload

AlexPeitz:
Add 128 bit UUID custom descriptors to noble GATT. Fixing issue #724.


jacobrosenthal:
Could use some cleanup, but also can you link any devices that use a 128bit descriptor?

AlexPeitz:
Yes it surely can, I don't understand why the checks fail for example.
I don't have a concrete Case of existig Devices with a 128 Bit UUID, but I'm using these for Research purposes and trying out new types of Descriptors. That's why I needed this fix.

rzr:
Please forward the change to:

PR #743, opened 2018-01-29
Fix an issue where Noble lazy inits before any event listeners can be…

### LINKED ISSUE #742: Noble lazy inits before any event listeners can be added.
Noble lazy inits before any event listeners can be added, leading to the `stateChanged` `poweredOff` event firing before any event listeners can catch it.

M1ras:
… added.  
Fixes #742.


sandeepmistry:
@M1ras thanks!

PR #730, opened 2018-01-06
Don't forget previously discovered characteristics

elafargue:
... when calling discoverCharacteristics multiple times in a program.

Without this fix, we get crashes when unsubscribing, and never receive 'data' events on previously subscribed characteristics if calling discoverCharacteristics multiple times with different uuid lists.

PR #747, opened 2018-01-31
Don't forget previously discovered characteristics

elafargue:
... when calling discoverCharacteristics multiple times in a program.

Without this fix, we get crashes when unsubscribing, and never receive 'data' events on previously subscribed characteristics if calling discoverCharacteristics multiple times with different uuid lists.

This PR includes comments/feedback from #730

### DIFF
--- a/lib/hci-socket/gatt.js
+++ b/lib/hci-socket/gatt.js
@@ -410,7 +410,7 @@ Gatt.prototype.discoverCharacteristics = function(serviceUuid, characteristicUui
   var service = this._services[serviceUuid];
   var characteristics = [];
 
-  this._characteristics[serviceUuid] = {};
+  this._characteristics[serviceUuid] = this._characteristics[serviceUuid] || {};
   this._descriptors[serviceUuid] = this._descriptors[serviceUuid] || {};
 
   var callback = function(data) {
--- a/lib/mac/highsierra.js
+++ b/lib/mac/highsierra.js
@@ -317,7 +317,7 @@ nobleBindings.on('kCBMsgId72', function(args) {
   var deviceUuid = args.kCBMsgArgDeviceUUID.toString('hex');
   var serviceUuids = [];
 
-  this._peripherals[deviceUuid].services = {};
+  this._peripherals[deviceUuid].services = this._peripherals[deviceUuid].services || {};
 
   if (args.kCBMsgArgServices) {
     for(var i = 0; i < args.kCBMsgArgServices.length; i++) {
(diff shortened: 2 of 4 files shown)


elafargue:
These are my changes. I will include follow-up commits that roll in the fixes described in the suggestions on #730 once I have tested those.

elafargue:
@jacobrosenthal tested this overnight with multiple connects/disconnects and discoveries, as well as multiple devices connected at the same time, all working fine. This removes a lot of irritating crashes that occur when using noble intensively.

jacobrosenthal:
What about that early exit PR https://github.com/sandeepmistry/noble/pull/383/files , that seemed kind of a classy way to do it. Can you run that and see if it similarly works well?

elafargue:
#383 won't work on its own, because it doesn't address the fact the bindings are squashing the references to previously discovered - and potentially already subscribed - services/characteristics. I can look at adding the early exit pattern of that PR if you want, it should speed things up a little bit. I'll try either way and report.

elafargue:
OK, just checked, with #383 on its own, the code crashes again at it did earlier - that PR might have solved edge cases, but it doesn't seem to provide a complete solution like this PR does.

Let me know if you can try this PR on your side and test on your side, and whether it works for you, merging it would be a tremendous help.

elafargue:
Did a few more tests - multiple connect/disconnect, multiple "discover" for services and characteristics in each session, this PR solves all the crashes or hangs I got earlier whereas the simple fix to "noble.js" in #383 really does not, I'm afraid.

sandeepmistry:
@elafargue thanks for updating, this looks good visually to me.

We'll need to do some testing here before merging.

jacobrosenthal:
Ill take a look with some of my existing code. @monteslu do we need anything for webble?

@tripish @Chondrules does this hit your needs from your previous PR?

monteslu:
I think we're good on the web ble bindings.  The call to `service.getCharacteristics()` seems to cache the characteristics internally in the web bluetooth api.  I've called it manually several times and get the same characteristics back.

jacobrosenthal:
Last attempt to try to ping @tripish @Chondrules to see if this solves their old PRs #383 and #541

BrantRoom3327:
Sorry @jacobrosenthal don't have any time to test your patch.  Gave it a glance and it seems like it could be a fix for the issue I was seeing.  If I get a chance to later I will but don't let me hold up anything.

elafargue:
Thank you!

PR #750, opened 2018-02-06
Websocket bugfix

hadrienk:
I tried to use the websocket bindings. This PR fixes the problems i faced: 

- Ability to set the host to connect to using NOBLE_WEBSOCKET_HOST.
- Send the `poweredOn` if it arrived before the websocket is setup.
- Send `poweredOff` upon socket errors.


sandeepmistry:
@hadrienk thanks.

PR #769, opened 2018-03-12
Fix multiple init of bindings

### LINKED ISSUE #765: Bindings being initialized twice or more
What title says.

This can be reproduced with the following

```javascript
const noble = require('noble');
noble.on('stateChange', () => {});
noble.on('stateChange', () => {}); 
x = noble.state;
```

I came across this bug when trying out [node-web-bluetooth](https://github.com/IjzerenHein/node-web-bluetooth), and using `btmon` I noticed that ATT MTU Exchange Requests were being sent 4 times when a connection was established. This was because listeners had been attached twice, and the initial MTU Exchange was called on a "chain" of 2 events.

I traced it to the lines 51-72 of `lib/noble.js`:
```javascript
  //lazy init bindings on first new listener, should be on stateChange
  this.on('newListener', function(event) {
    if (event === 'stateChange' && !this.initialized) {
      process.nextTick(function() {
        this._bindings.init();
        this.initialized = true;
      }.bind(this));
    }
  }.bind(this));

  //or lazy init bindings if someone attempts to get state first
  Object.defineProperties(this, {
    state: {
      get: function () {
        if (!this.initialized) {
          this._bindings.init();
          this.initialized = true;
        }
        return this._state;
      }
    }
  });
```

So you can have initialization queued with `process.nextTick`, but `initialized` remains false. Until the initializer actually gets called you can queue any number of other initializations with `.on('stateChange',func)`, which will be later executed. You can also initialize by accessing `noble.state`, which will set `initialized` to true immediately, but will not prevent previously queued initializations from executing.

Edit: I should say this issue is due to changes in version 1.9.0, so I went back to 1.8.1 for now.

sandeepmistry:
By setting the initialized flag to true before calling init.

Fixes #765.

### DIFF
--- a/lib/noble.js
+++ b/lib/noble.js
@@ -51,9 +51,10 @@ function Noble(bindings) {
   //lazy init bindings on first new listener, should be on stateChange
   this.on('newListener', function(event) {
     if (event === 'stateChange' && !this.initialized) {
+      this.initialized = true;
+
       process.nextTick(function() {
         this._bindings.init();
-        this.initialized = true;
       }.bind(this));
     }
   }.bind(this));
@@ -63,8 +64,9 @@ function Noble(bindings) {
     state: {
       get: function () {
         if (!this.initialized) {
-          this._bindings.init();
           this.initialized = true;
+
+          this._bindings.init();
         }
         return this._state;
       }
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
Works on high sierra. Lets hope this gets us there finally :)

piersbarrios:
Works perfectly on RaspberryPi 3 (tell me if you want more details)

PR #772, opened 2018-03-16
Increase scan interval and window to 100 ms

brghena:
Scan interval defines the frequency that the scanner hops between advertising channels. Periodically switching has the advantage of avoiding possible interference on those channels.

However, in practice scanning hardware (like a BLE dongle) have blind regions when switching channels where they cannot receive packets. According to [a recent paper](https://pdfs.semanticscholar.org/7cc4/ec16353c9191e9c7e65bb3d80aa4df204ff6.pdf), these blind regions can be as long as milliseconds (depending on hardware and configurations). So, increasing the time between scanner frequency hops should result in an increased advertisement reception rate.

Adjusting the scan interval and window to 100 ms is an arbitrary choice. It felt sufficiently high, without lingering on a channel for too long.

This pull request was tested by running `examples/advertisement-discovery.js` for several seconds to check that BLE advertisements are still being discovered.

### DIFF
--- a/lib/hci-socket/hci.js
+++ b/lib/hci-socket/hci.js
@@ -264,8 +264,8 @@ Hci.prototype.setScanParameters = function() {
 
   // data
   cmd.writeUInt8(0x01, 4); // type: 0 -> passive, 1 -> active
-  cmd.writeUInt16LE(0x0010, 5); // internal, ms * 1.6
-  cmd.writeUInt16LE(0x0010, 7); // window, ms * 1.6
+  cmd.writeUInt16LE(0x00A0, 5); // scan interval, in units of 0.625 ms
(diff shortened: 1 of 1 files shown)


ppannuto:
Came across something to quantify this a bit today in other research. Looks like the nrf family [takes ~800 us to switch listening channels](https://devzone.nordicsemi.com/f/nordic-q-a/25553/nrf52832-scan-channel-switch-time).

For a 20 ms scan interval, this means that almost 5% of the time the radio isn't listening at all. Upping to a 100 ms interval would drop this to 1%.

ghost:
https://stackoverflow.com/questions/37307301/ble-scan-interval-windows-10/46653226#46653226

So for windows: 

> The scan parameters are hardcoded to a scan interval of 118.125 ms and scan window of 18.125 ms.

rzr:
Please also check with:

PR #789, opened 2018-04-30
Listener option

jszersze:
Added optional listeners on connection and services to remove a memory leak when connecting/disconnection multiple times.


rzr:
Is there any traction for this change ?

PR #794, opened 2018-05-10
Lowercasization of the Bluetooth Address

ouinouin:
Hi, here is the pull request to force the bluetooth mac address sent back by  the bluetooth processLeAdvertisingReport scan function to be in lower case.
( see https://github.com/noble/noble-device/pull/30 )


don:
toLowerCase() in this code doesn't do anything. The Buffer.toString('hex') returns lower case.

    > b = new Buffer([15, 14, 13, 12, 11, 10])
    <Buffer 0f 0e 0d 0c 0b 0a>
    > b.toString('hex').match(/.{1,2}/g).reverse().join(':')
    '0a:0b:0c:0d:0e:0f'

PR #800, opened 2018-06-05
fix npm audit warnings

don:
Note that there is still a low severity vulnerability from jshint > lodash https://nodesecurity.io/advisories/577 jshint has fixed this, so it will be available in a future release jshint/jshint@e0b4e91

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   },
   "devDependencies": {
     "async": "~0.2.9",
+    "jshint": "latest",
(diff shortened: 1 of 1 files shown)


jrobeson:
Removal of previous versions of node requires explicit approval and is done separately here https://github.com/noble/noble/pull/691

rzr:
May I suggest to submit your patches to:
https://github.com/abandonware/noble

PR #802, opened 2018-06-08
connect to bleno echo service example

don:


don:
CI failed on Node 0.8 unrelated to this example

PR #803, opened 2018-06-08
Fixes GATT bug when calling characteristic.subscribe() that caused RangeError exception

pcodes:
Before, the GATT code expected that messages from a ble device for `subscribe()` would be long enough to read 4 bytes from. However, I ran into a problem where when a message was type 2 from my device, it did not have enough data to read from and thus the program would crash with a `RangeError` exception . I admittedly do not entirely understand what the types are supposed to represent, especially as the `type` variable is not used elsewhere in the code block. Either way, this change fixes the problem I encountered.


rzr:

PR #813, opened 2018-07-03
OSX fix custom service 128 bit UUID to 16 bit

yackermann:
So  currently OSX when it doesn't understand service UUID, such as FIDO custom uuid is fffd, it returns full 128 bit UUID(0000fffd00001000800000805f9b34fb). However on windows it always returns 16bit UUID i.e fffd

This change fixes OSX bindings to ensure that for any service, 16bit UUID is always returned.

### DIFF
--- a/lib/mac/highsierra.js
+++ b/lib/mac/highsierra.js
@@ -322,8 +322,9 @@ nobleBindings.on('kCBMsgId72', function(args) {
   if (args.kCBMsgArgServices) {
     for(var i = 0; i < args.kCBMsgArgServices.length; i++) {
       var serviceUuid = args.kCBMsgArgServices[i].kCBMsgArgUUID.toString('hex');
-      if(serviceUuid.length > 4)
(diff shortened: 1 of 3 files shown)


yackermann:
P.S. I think you guys need to remove tests for node 0.8 as TLS directory for it long expired

lib/mac/highsierra.js
```diff
@@ -321,8 +321,13 @@ nobleBindings.on('kCBMsgId72', function(args) {
 
   if (args.kCBMsgArgServices) {
     for(var i = 0; i < args.kCBMsgArgServices.length; i++) {
+      var discoveredServiceUuid = args.kCBMsgArgServices[i].kCBMsgArgUUID.toString('hex');
+      if(discoveredServiceUuid.length > 4) {
+        discoveredServiceUuid = discoveredServiceUuid.slice(4, 8);
+      }
+
```
ghost, line 328: seems like this should be a utility function rather than copy/pasted in all the places you're using it.

Issue #864, opened 2019-02-18
LE connection failure with "Error: Command Disallowed"


abhishekkumardwivedi:
The same piece of code works in my laptop with ubuntu but fails in another Debian based arm device. I tried with peripheral explorer code which also failed with same issue. Following are the error log from hcidump:
```
< HCI Command: Set Event Mask (0x03|0x0001) plen 8
    Mask: 0xfffffbff07f8bf3d
> HCI Event: Command Complete (0x0e) plen 4
    Set Event Mask (0x03|0x0001) ncmd 1
    status 0x00
< HCI Command: LE Set Event Mask (0x08|0x0001) plen 8
    mask 0x1f00000000000000 (Reserved)
> HCI Event: Command Complete (0x0e) plen 4
    LE Set Event Mask (0x08|0x0001) ncmd 1
    status 0x00
< HCI Command: Read Local Version Information (0x04|0x0001) plen 0
> HCI Event: Command Complete (0x0e) plen 12
    Read Local Version Information (0x04|0x0001) ncmd 1
    status 0x00
    HCI Version: 4.0 (0x6) HCI Revision: 0x1000
    LMP Version: 4.0 (0x6) LMP Subversion: 0x220e
    Manufacturer: Broadcom Corporation (15)
< HCI Command: Write LE Host Supported (0x03|0x006d) plen 2
  0000: 01 00                                             ..
> HCI Event: Command Complete (0x0e) plen 4
    Write LE Host Supported (0x03|0x006d) ncmd 1
    0000: 00                                                .
< HCI Command: Read LE Host Supported (0x03|0x006c) plen 0
> HCI Event: Command Complete (0x0e) plen 6
    Read LE Host Supported (0x03|0x006c) ncmd 1
    0000: 00 01 00                                          ...
< HCI Command: Read BD ADDR (0x04|0x0009) plen 0
> HCI Event: Command Complete (0x0e) plen 10
    Read BD ADDR (0x04|0x0009) ncmd 1
    status 0x00 bdaddr 06:13:05:02:12:31
< HCI Command: LE Set Scan Enable (0x08|0x000c) plen 2
    value 0x00 (scanning disabled)
    filter duplicates 0x01 (enabled)
> HCI Event: Command Complete (0x0e) plen 4
    LE Set Scan Enable (0x08|0x000c) ncmd 1
    status 0x0c
    Error: Command Disallowed
< HCI Command: LE Set Scan Parameters (0x08|0x000b) plen 7
[...]

PR #865, opened 2019-02-22
open up reset API for application

abhishekkumardwivedi:
Resetting device before connect solves the issue #864 .

The device I am using is having Broadcom serial bluetooth module. After debugging the issue, I could find, there might not be a problem from noble connection command. But this could be the issue at module firmware. As "Command Disallowed" error actually comes from the module firmware to the kernel to userspace and so to noble via hci socket. I did verify this issue by hcitool also. And before connect, we can do reset the bluetooth device to overcome this issue. This is a workaround, and I feel that actual fix should be in Bluetooth device firmware. But, we can't expect everyone will have very updated bluetooth firmware as well as the latest kernel.

So to have reset command available to the application gives a wide set of noble users an option to overcome with this kind of issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,14 @@ __NOTE:__ ```noble.state``` must be ```poweredOn``` before scanning is started. 
 noble.stopScanning();
 ```
 
+#### Reset device
+
+```javascript
+noble.reset();
+```
+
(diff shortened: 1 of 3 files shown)


RickVM (approved):

somq:
Good PR, thanks. Should be merged!

rzr:

PR #882, opened 2019-05-14
allow connection to all `connectable` peripherals

adamlouis:
Allow connection to any peripheral where `peripheral.connectable` is `true`.

`peripheral.state` can become out of sync with `peripheral.connectable` in cases where there is no disconnect event.

This happens when plugging / unplugging an external bluetooth adapter on MacOS while a peripheral is connected as described here: https://github.com/noble/noble/issues/881


rzr:

PR #889, opened 2019-06-17
Added MTU value to peripheral on successful handshake

EricDobyns:
Added a quick update to retrieve the MTU value for a specific peripheral. Majority of the code was already there but for whatever reason the NobleBindings.prototype.onMtu method was left blank 🤔


rzr:
Please confirm those changes are working well in latest @abadonware/noble release

PR #890, opened 2019-06-19
switch to xpc-connect to support Mojave

jongear:
`xpc-connect` is a fork of `xpc-connection` with support for Mojave.


zbot473:
Why was this closed?

jhackett1:
is this the cause of https://github.com/noble/noble/issues/980?

PR #911, opened 2019-10-22
Patch Error Ws

Floriantoine:
To avoid bugs when implementing websocket with the use of overlay like quasar

to ```node_modules/noble/lib/websocket/bindings.js```
change ```var WebSocket = require('ws');``` -> ```var WebSocket = window.require('ws');```


rzr:

PR #934, opened 2020-04-16
This location must be true when using noble-mac on mojyave or later OS

k-katsuda:
When using noble-mac, localName could be obtained by applying the modified part


rzr:
May this be forwarded to @abandonware

PR #985, opened 2022-02-18
Support the HomeBridge Docker instalation - Core bluetooth

wmar:
Does the product support HomeBridge in docker version?  I would like to try in the Synology NAS with HomeBridge and MI Bluetooth humidifier & temperature sensor product.


rzr:
Please forward downstream too:

https://github.com/abandonware/noble/issues/2

